PR #23, opened 2018-10-16
E-resources: Update paths and list agreements

doytch:
- Updated the E-resources section to the new `/erm/resource/*` paths.
- Fixed a paging bug.
  - Query params specify `perPage` again and use the `offset` param rather than the `page` param.
- In the E-resource details pane, show the agreements for the selected e-resource.
  - Using the new `/erm/resource/{id}/entitlements` path
  - Questions for @sosguthorpe:
    - Do you know if **Acquisition method** should be `entitlement.resource.class === 'org.olf.kb.Pkg' ? 'Package' : 'Item'` [for/based on Scenario 2 here](https://issues.folio.org/browse/UIER-21).
    - Some items returned from `/erm/resource/electronic` return a 404 when I look up their `/entitlements`. Should they be returning an empty array? E.g., the "American Psychological Association:Master" resource. Are they just anomalies of sample data?

### DIFF
--- a/src/ERM.js
+++ b/src/ERM.js
@@ -7,10 +7,12 @@ import Agreements from './routes/Agreements';
 import EResources from './routes/EResources';
 import Tabs from './components/Tabs';
 
+const INITIAL_RESULT_COUNT = 100;
+
 export default class Erm extends React.Component {
   static manifest = Object.freeze({
     query: {},
-    resultCount: { initialValue: 0 }
+    resultCount: { initialValue: INITIAL_RESULT_COUNT }
   });
 
   static propTypes = {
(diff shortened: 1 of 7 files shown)


sosguthorpe:
> * Do you know if **Acquisition method** should be `entitlement.resource.class === 'org.olf.kb.Pkg' ? 'Package' : 'Item'`

`entitlement.resource.class === 'org.olf.kb.Pkg' ? 'Package' : 'Title' ` works fine if we are only showing 2 vals yes.
`entitlement.resource.class === 'org.olf.kb.Pkg' ? 'Package' : entitlement.resource.class === 'org.olf.kb.PackageContentItem' ? 'Package Title' : 'Title'` gives more detail.

> * Some items returned from `/erm/resource/electronic` return a 404 when I look up their `/entitlements`. Should they be returning an empty array? E.g., the "American Psychological Association:Master" resource. Are they just anomalies of sample data?

Hmm.. I think I tend to return a 404 when an ID comes in for an object type we weren't expecting. I'll take a look at this as this might be a bug. Semantics between 'there are no paths to get this resource' and 'it is impossible to link this resource to an entitlement' are different.

PR #33, opened 2018-11-15
add acorn to dependencies to stop npm complaining

aclseelen:


aclseelen:
Thanks for your review Mark. The warning comes from `npm`. I will just try to ignore it, but what could happen is that other developers will (by this warning) be tempted to install acorn as well and have their `package.json` edited, which is kind of annoying when working with git and you can't really exclude that file.

doytch:
@aclseelen Well we recommend using `yarn` rather than `npm` for a few reasons, and we can chalk this up as just another reason. Like I say, I've never seen this warning using yarn.

PR #46, opened 2018-12-06
Organization Selection with an async options list

doytch:
**Problem**
Users can add organizations to an agreement. They do so by selecting which organization they want to add to the agreement from a `Selection`'s dropdown. The dropdown's option list was being populated by all the records at the `/erm/org` endpoint. However, that endpoint can return thousands of entries. We want to be able to search that endpoint based on the user's entered text so we can do filtering on the backend rather than locally.

**Solution**
- Create a component `OrganizationSelection` that renders it's descendants `OrganizationSelectionContainer` > `OrganizationSelectionDisplay`.
- `OS` merely `connect()`s `OSC` with a unique `dataKey` generated from the form field's name. This is so we can fetch different sets of orgs for each dropdown.
- `OSC` handles fetching of `/erm/org?our_query`, and translation of the resources to its child, `OSD`. 
  - One strange part is that it always `unshift`s the currently-selected option (not just currently-selected _value_) into the orgs lists that was fetched so that it may be rendered. Which naturally means I need to hold the currently-selected option in `state` manually.
- `OSD` renders the `Selection` itself.

**Note to Reviewers**
I have a feeling we're going to start having these use cases crop up more as we start getting real data and reckoning with the fact that we haven't been assuming large enough datasets. At least, I know _I'm_ going to facing these kinds of scenarios going forward with ERM. So two things are: 
- We might want to consider whether we want to deal with these scenarios manually like I did here or provide some support via Stripes.
- If y'all have a few spare moments, I'd appreciate your peek at it to see if there'd be a more elegant/performant/succinct solution to what I've got going here.

**Screencast:** [imgur gif](https://imgur.com/a/xQzHGIA) 
(Ignore the duplicate entries in the dropdown, it was me messing around creating a bunch of duplicates)

### DIFF
--- a/src/components/Agreements/AgreementForm/Sections/AgreementFormOrganizations.js
+++ b/src/components/Agreements/AgreementForm/Sections/AgreementFormOrganizations.js
@@ -83,7 +83,6 @@ class AgreementFormOrganizations extends React.Component {
                 <Field
                   component={OrganizationSelection}
                   name={`orgs[${index}].org`}
-                  stripes={this.props.stripes}
                 />
               </Col>
               <Col xs={3}>
--- a/src/components/Agreements/AgreementForm/Sections/AgreementFormOrganizations.js
+++ b/src/components/Agreements/AgreementForm/Sections/AgreementFormOrganizations.js
@@ -23,7 +23,7 @@ class AgreementFormOrganizations extends React.Component {
     onToggle: PropTypes.func,
     open: PropTypes.bool,
     parentResources: PropTypes.shape({
-      orgs: PropTypes.object,
+      orgRoleValues: PropTypes.object,
     }),
     stripes: PropTypes.shape({
       connect: PropTypes.func,
--- a/src/components/Agreements/AgreementForm/Sections/AgreementFormOrganizations.js
+++ b/src/components/Agreements/AgreementForm/Sections/AgreementFormOrganizations.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
-import { debounce, get } from 'lodash';
+import { get } from 'lodash';
 import { Field, FieldArray } from 'redux-form';
 
 import { withStripes } from '@folio/stripes/core';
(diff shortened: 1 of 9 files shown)


src/components/OrganizationSelection/OrganizationSelectionContainer.js
```diff
@@ -0,0 +1,91 @@
...
+export default class OrganizationSelectionContainer extends React.Component {
+  static manifest = Object.freeze({
+    organizations: {
+      type: 'okapi',
+      path: 'erm/org',
+      limitParam: 'perPage',
+      perRequest: 100,
+      params: {
+        match: 'name',
+        term: '${orgNameFilter}', // eslint-disable-line no-template-curly-in-string
```
zburke: Use `%{foo}` to substitute in a local resource's value. `${foo}` [has been deprecated](https://github.com/folio-org/stripes-connect/blob/master/doc/api.md#text-substitution) for a long time, and eslint doesn't get offended either.

src/routes/Agreements.js
```diff
@@ -50,12 +50,11 @@ class Agreements extends React.Component {
...
       type: 'okapi',
       path: 'erm/org',
-      records: 'results',
       limitParam: 'perPage',
       perRequest: 100,
-      recordsRequired: '1000',
       params: {
-        stats: 'true',
+        match: 'name',
+        term: '${orgNameFilter}', // eslint-disable-line no-template-curly-in-string
```
zburke: Same here: `%{orgNameFilter}` makes the same substitution as `${orgNameFilter}` and avoids the wrath of the eslint gods.

src/components/OrganizationSelection/OrganizationSelectionContainer.js
doytch: Oh my god I just realised that I've read that section of the API as the direct opposite of what it actually says so many times...

src/components/Agreements/AgreementForm/Sections/AgreementFormOrganizations.js
```diff
@@ -90,24 +80,11 @@ class AgreementFormOrganizations extends React.Component {
...
-                        });
-                      }}
-                      placeholder={placeholder}
-                    />
-                  )}
-                </FormattedMessage>
+                <Field
+                  component={OrganizationSelection}
+                  name={`orgs[${index}].org`}
+                  stripes={this.props.stripes}
```
cherewaty: I've been preferring a pattern of using `withStripes()` inside `OrganizationSelection` instead of prop-drilling the `stripes` god object down into children.

PR #114, opened 2019-02-21
Erm 35

CalamityC:
(1) "Filter by Organisation-Role does not change default Agreement Search filters":
the preset status filters are deleted when an organisation is selected, if manually set the other filters remain
(2) "pre-definded roles":
is this a backend issue? vendor role is still not present and more roles than the pre-defined are shown

### DIFF
--- a/src/components/Agreements/AgreementFilters/AgreementFilters.js
+++ b/src/components/Agreements/AgreementFilters/AgreementFilters.js
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import { get } from 'lodash';
 import { FormattedMessage } from 'react-intl';
 
-import { Accordion, AccordionSet, FilterAccordionHeader } from '@folio/stripes/components';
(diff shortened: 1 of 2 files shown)


CalamityC:
Hi Mark, I can do that but not before Monday, I‘m on my way to Amsterdam
________________________________________
Von: Mark Deutsch <notifications@github.com>
Gesendet: Donnerstag, 21. Februar 2019 21:11:03
An: folio-org/ui-agreements
Cc: Malzer, Claudia; Mention
Betreff: Re: [folio-org/ui-agreements] Erm 35 (#114)

@doytch approved this pull request.

@CalamityC<https://github.com/CalamityC> can you file a bug about the first point you raised regarding the default filters getting cleared, and assign it to me? I noticed that bug and started going down the debugging for that but it's not a simple fix so I shelved it.

Regarding the second point, yeah that'll get fixed by the backend being updated to include folio-org/mod-agreements#86<https://github.com/folio-org/mod-agreements/pull/86>

PR #130, opened 2019-03-08
 UIER-32: Multiple financial agreement lines

julianladisch:


src/components/Agreements/ViewAgreement/Sections/Finances.js
```diff
@@ -0,0 +1,62 @@
...
+import { get } from 'lodash';
+import { FormattedMessage } from 'react-intl';
+import { Accordion, Badge, Button, Icon } from '@folio/stripes/components';
+
+import FinancesAgreementLines from './FinancesAgreementLines';
+
+export default class Finances extends React.Component {
+  static propTypes = {
+    financesAgreementLines: PropTypes.arrayOf(PropTypes.object),
+    fetchMoreEresources: PropTypes.func,
```
doytch: Can you remove this from the propTypes? It doesn't look like you're using or passing it in.

src/components/Agreements/ViewAgreement/Sections/Finances.js
```diff
@@ -0,0 +1,62 @@
...
+        {this.renderBadge()}
+      </div>;
+
+    return (
+      <Accordion
+        id={id}
+        label={<FormattedMessage id="ui-agreements.agreements.financesAgreementLines" />}
+        open={open}
+        onToggle={onToggle}
+        displayWhenClosed={buttonAndBadge}
```
doytch: Are you sure we're rendering both the button and badge in both scenarios (open/closed)? I was under the impression that we render button when open and badge when closed. Gill can probably confirm one way or another.

julianladisch: In this pull request (UIER-32) I did not change the positioning and visibility of the edit button. That was done in https://issues.folio.org/browse/UIER-31 . I only moved the button code from FinancesAgreementLines.js to Finances.js.

julianladisch: If it is an issue it can be reported in UIER-31 or in a separate jira ticket.

julianladisch: Files as https://issues.folio.org/browse/UIER-54

PR #133, opened 2019-03-08
ERM-30: Don't throw errors from orders-storage endpoint

doytch:
We should come up with a longer-term solution than this but as-is `ui-agreements` is broken if run on a platform without whatever module implements the `orders-storage` endpoint.

### DIFF
--- a/src/components/Agreements/ViewAgreement/ViewAgreement.js
+++ b/src/components/Agreements/ViewAgreement/ViewAgreement.js
@@ -71,6 +71,7 @@ class ViewAgreement extends React.Component {
       },
       records: 'po_lines',
       shouldRefresh,
+      throwErrors: false,
     },
     agreementEresources: {
       type: 'okapi',


src/components/Agreements/ViewAgreement/ViewAgreement.js
```diff
@@ -71,6 +71,7 @@ class ViewAgreement extends React.Component {
       },
       records: 'po_lines',
       shouldRefresh,
+      throwErrors: false,
```
aliaksei-chumakou: @doytch Are you sure that calling `orders-storage` is a good idea? Could you please check mod-orders API? https://s3.amazonaws.com/foliodocs/api/mod-orders/order.html#orders_order_lines_get
AFAIK, we agreed that `mod-orders-storage` is kind of internal, and clients reach `mod-orders` API
cc @craigmcnally

craigmcnally: That's right @aliaksei-chumakou.  There is a lot of business logic in mod-orders that will be bypassed if talking directly to the storage layer.  This is more important for create/update/delete operations, but also includes get operations as well.  For instance, summary information is calculated in the business logic module (mod-orders).  

@doytch we should touch base and discuss your use-case.  Let me know what time works for you.

doytch: Appreciate it @craigmcnally! 

Could you actually link up with Julian, who’s working on this functionality as part of this PR: https://github.com/folio-org/ui-agreements/pull/130

PR #141, opened 2019-03-19
UIER-34: Display "show invoices" link

julianladisch:
If there exists an invoice line where one of the poLineIds matches one of the poLineIds of the
current agreement then display the "Show invoices" link.

Endpoint to query: /invoice-storage/invoice-lines
https://github.com/folio-org/mod-invoice-storage/blob/master/ramls/invoice.raml
https://github.com/folio-org/acq-models/blob/master/mod-invoice-storage/schemas/invoice_line.json

The link label is shown but without a clickable link because the destination doesn't exist yet:
https://github.com/folio-org/ui-invoice

### DIFF
--- a/src/components/Agreements/ViewAgreement/Sections/Finances.js
+++ b/src/components/Agreements/ViewAgreement/Sections/Finances.js
@@ -10,6 +10,7 @@ export default class Finances extends React.Component {
   static propTypes = {
     financesAgreementLines: PropTypes.arrayOf(PropTypes.object),
     id: PropTypes.string,
+    invoices: PropTypes.arrayOf(PropTypes.object),
     onToggle: PropTypes.func,
     open: PropTypes.bool,
   };
@@ -20,6 +21,14 @@ export default class Finances extends React.Component {
       return <Icon icon="spinner-ellipsis" width="10px" />;
     }
     return <Badge data-test-finances-agreement-lines-count={count}>{count}</Badge>;
+  }
+
+  renderInvoices = () => {
+    const invoices = get(this.props, 'invoices', []);
+    if (!invoices.length) {
+      return null;
+    }
+    return <div align="end"><FormattedMessage id="ui-agreements.agreements.showInvoicesLink" /></div>;
   }
 
   render() {
@@ -45,6 +54,7 @@ export default class Finances extends React.Component {
         displayWhenClosed={buttonAndBadge}
         displayWhenOpen={buttonAndBadge}
       >
+        {this.renderInvoices()}
         <FinancesAgreementLines
           visible={open}
           {...this.props}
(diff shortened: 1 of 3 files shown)


src/components/Agreements/ViewAgreement/ViewAgreement.js
```diff
@@ -73,6 +73,14 @@ class ViewAgreement extends React.Component {
       shouldRefresh,
       throwErrors: false,
     },
+    invoices: {
+      type: 'okapi',
+      path: 'invoice-storage/invoice-lines',
```
doytch: Can you confirm that you've checked with their team about this path? They pushed back against us accessing `orders-storage` so I don't want to merge another bit of code that directly accesses their storage modules.

src/components/Agreements/ViewAgreement/ViewAgreement.js
```diff
@@ -135,6 +144,7 @@ class ViewAgreement extends React.Component {
 
     if ((prevAgreement.id !== agreement.id) || (difference(users, prevUsers).length)) {
       this.fetchUsers();
+      this.fetchInvoices();
```
doytch: Can we move this into it's own condition since it doesn't appear to have anything to do with contacts?

Correct me if I'm wrong, but I'm thinking we want to refetch when either:
- a new agreement has been selected (`id`s are different)
- the number of `financesAgreementLines` is different

src/components/Agreements/ViewAgreement/Sections/Finances.js
```diff
@@ -22,6 +23,14 @@ export default class Finances extends React.Component {
     return <Badge data-test-finances-agreement-lines-count={count}>{count}</Badge>;
   }
 
+  renderInvoices = () => {
+    const invoices = get(this.props, 'invoices', []);
+    if (!invoices.length) {
+      return null;
+    }
+    return <div align="end"><FormattedMessage id="ui-agreements.agreements.showInvoicesLink" /></div>;
```
doytch: Can you use the `Layout` component directly here instead of rendering `div` markup?

Here's its [readme](https://github.com/folio-org/stripes-components/tree/master/lib/Layout) and [storybook](https://ux.folio.org/storybook/?selectedKind=Layout&selectedStory=Available%20classes&full=0&addons=1&stories=1&panelRight=0&addonPanel=REACT_STORYBOOK%2Freadme%2Fpanel)

src/components/Agreements/ViewAgreement/ViewAgreement.js
julianladisch: Until now mod-invoice is "Not supported":
https://github.com/folio-org/mod-invoice/blob/master/src/main/java/org/folio/rest/impl/InvoicesImpl.java#L23

src/components/Agreements/ViewAgreement/ViewAgreement.js
julianladisch: Thanks for catching this wrong check; changed.

src/components/Agreements/ViewAgreement/Sections/Finances.js
julianladisch: Yes, done.

julianladisch:
Added hasInterface check.

PR #155, opened 2019-03-25
Erm-98: Rendering controlling license' terms using new component

CalamityC:
modify routes/Agreements.js to fetch license terms
modify LicenseBusinessTerms.js to use LicenseTermsList component
modify package.json to use stripes-erm-components 1.2.1
add license.noControllingLicense to en.json
add test license-tems


src/components/Agreements/ViewAgreement/Sections/LicenseBusinessTerms.js
```diff
@@ -1,25 +1,53 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { get } from 'lodash';
 import { FormattedMessage } from 'react-intl';
 import { Accordion } from '@folio/stripes/components';
+import { LicenseTermsList } from '@folio/stripes-erm-components';
 
 export default class LicenseBusinessTerms extends React.Component {
   static propTypes = {
+    linkedLicenses: PropTypes.shape({
```
doytch: `linkedLicenses` is inside `agreement`, as that's where we pull it from on L36. This should be something like:
```
agreement: PropTypes.shape({
  linkedLicenses: PropTypes.arrayOf(PropTypes.shape({
    remoteId_object: ...
    ... 
  }))
}
```

Let's get this fixed since as-is, it's suggesting something incorrect to us and other developers.

(While we're at it, let's alpha-sort the `linkedLicenses` properties. Ie, `remoteId_object` before `status`)

src/components/Agreements/ViewAgreement/Sections/LicenseBusinessTerms.js
```diff
@@ -1,25 +1,53 @@
...
+     <Accordion
+       id={this.props.id}
+       label={<FormattedMessage id="ui-agreements.agreements.licenseAndBusTerms" />}
+       open={this.props.open}
+       onToggle={this.props.onToggle}
+     >
+       { controllingLicense ?
+         this.renderLicenseTermsList(controllingLicense) :
+         <FormattedMessage id="ui-agreements.license.noControllingLicense" />
+         }
```
doytch: Closing brace should be in the same column as the opening brace. I'm not sure why the linter isn't catching this (it doesn't even throw an error when I checked out this branch) - it should be 😬

src/routes/Agreements.js
```diff
@@ -45,6 +45,10 @@ class Agreements extends React.Component {
       path: 'erm/sas/${selectedAgreementId}', // eslint-disable-line no-template-curly-in-string
       fetch: false,
     },
+    terms: {
+      type: 'okapi',
+      path: 'licenses/custprops',
```
doytch: So we're fetching from one of the `/licenses` endpoints here, which is something we haven't done before. The `licenses` endpoints are provided by `mod-licenses`, which provides them as part of the `licenses` **Okapi interface**. 

Whenever we're depending on a backend interface in UI code, we need to express that dependency somewhere. Whenever a dependency is mandatory, we do it in our frontend module's `package.json`. If you look at `ui-licenses`'s package.json, you'll see there's an `okapiInterfaces` key in the `stripes` object that lists the dependency on licenses. 

So what we need to do is copy that key/value into the `ui-agreements`'s package.json (which already expresses its dependency on the `erm` interface) to show that we're depending on that interface as well.

PR #170, opened 2019-04-11
Add Tags Helper app to Agreements, refs erm-173

aditya-matukumalli:
- Used withTags HOC to render the tags app and all the CRUD operations
- Added Translations


src/components/Agreements/ViewAgreement/ViewAgreement.js
```diff
@@ -42,7 +45,10 @@ class ViewAgreement extends React.Component {
     selectedAgreement: {
       type: 'okapi',
       path: 'erm/sas/:{id}',
-      shouldRefresh,
+      shouldRefresh: (resource, action, refresh) => {
+        const { path } = action.meta;
+        return refresh || (path && path.match(/link/));
```
doytch: What's this do?

src/components/Agreements/ViewAgreement/ViewAgreement.js
```diff
@@ -376,6 +384,35 @@ class ViewAgreement extends React.Component {
...
   }
 
+  renderDetailMenu(agreement) {
+    const {
+      tagsEnabled,
+      tagsToggle,
+    } = this.props;
+
+
+    const tags = ((agreement && agreement.tags) || {}) || [];
```
doytch: This doesn't appear to make sense. The last false condition will never be executed.

Can you write these kinds of statements with `get` in the ERM apps instead? eg, `get(agreement, 'tags', [])`?

src/components/Agreements/ViewAgreement/ViewAgreement.js
```diff
@@ -376,6 +384,35 @@ class ViewAgreement extends React.Component {
...
+
+    return (
+      <PaneMenu>
+        {
+          tagsEnabled &&
+          <FormattedMessage id="ui-requests.showTags">
```
doytch: Don't depend on ui-requests for strings.

src/components/Agreements/ViewAgreement/ViewAgreement.js
```diff
@@ -465,4 +503,6 @@ class ViewAgreement extends React.Component {
   }
 }
 
-export default ViewAgreement;
+export default compose(
```
doytch: We don't really need `compose` here given that we're just using one HOC, right?

src/routes/Agreements.js
```diff
@@ -1,10 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { injectIntl, intlShape } from 'react-intl';
-
+import { FormattedDate, FormattedMessage } from 'react-intl';
 import { SearchAndSort } from '@folio/stripes/smart-components';
-import { getSASParams } from '@folio/stripes-erm-components';
-
+import { getSASParams, Tags } from '@folio/stripes-erm-components';
```
doytch: Make sure to update the version of stripes-erm-components you're relying on for the `Tags` component after you bump the version.

src/routes/Agreements.js
```diff
@@ -192,19 +190,19 @@ class Agreements extends React.Component {
...
         browseOnly={this.props.browseOnly}
         columnMapping={{
-          name: intl.formatMessage({ id: 'ui-agreements.agreements.name' }),
-          vendor: intl.formatMessage({ id: 'ui-agreements.agreements.vendorInfo.vendor' }),
-          startDate: intl.formatMessage({ id: 'ui-agreements.agreements.startDate' }),
-          endDate: intl.formatMessage({ id: 'ui-agreements.agreements.endDate' }),
-          cancellationDeadline: intl.formatMessage({ id: 'ui-agreements.agreements.cancellationDeadline' }),
-          agreementStatus: intl.formatMessage({ id: 'ui-agreements.agreements.agreementStatus' }),
-          lastUpdated: intl.formatMessage({ id: 'ui-agreements.lastUpdated' }),
+          name: <FormattedMessage id="ui-agreements.agreements.name" />,
```
doytch: Niiiice, thx!

test/ui-testing/tags-crud.js
```diff
@@ -0,0 +1,131 @@
+/* global describe, it, before, after, Nightmare */
+const AgreementCRUD = require('./agreement-crud');
+
+module.exports.test = (uiTestCtx) => {
+  const number = Math.round(Math.random() * 100000);
+  const testTag = `test${Math.floor(Math.random() * 10000)}`;
+  const agreement = {
+    name: `Linked License Agreement #${number}`,
```
doytch: Call it something like `Tags Agreement #${number}` instead. Helps to distinguish the agreements created after the fact.

test/ui-testing/tags-crud.js
```diff
@@ -0,0 +1,131 @@
...
+            done();
+          })
+          .catch(done);
+      });
+
+      it('should create agreement', done => {
+        AgreementCRUD.createAgreement(nightmare, done, agreement);
+      });
+
+      it(`should search for and find agreement: ${agreement.name}`, done => {
```
doytch: I've got no issues with you being extra cautious in tests and having specific cases for things that aren't integral to the _real_ thing we're testing (tags in this case), but I tend not to because:
* I'm lazy
* It's already tested in another place, eg `agreement-crud`
* It slows down running the tests

Up to you, just letting you know for future use.

test/ui-testing/tags-crud.js
```diff
@@ -0,0 +1,131 @@
...
+          .click(config.select.settings)
+          .wait('#app-list-item-clickable-settings')
+          .wait('a[href="/settings/tags"]')
+          .click('a[href="/settings/tags"]')
+          .wait('a[href="/settings/tags/general"]')
+          .click('a[href="/settings/tags/general"]')
+          .wait('#tags_enabled')
+          .wait(222)
+          .evaluate(() => {
+            const element = document.querySelector('[value="true"]');
```
doytch: This feels like a really broad selector to be using. I'm afraid of maintainability here the moment that Folio adds another setting to that page.

test/ui-testing/tags-crud.js
```diff
@@ -0,0 +1,131 @@
...
+      it('should create agreement', done => {
+        AgreementCRUD.createAgreement(nightmare, done, agreement);
+      });
+
+      it(`should search for and find agreement: ${agreement.name}`, done => {
+        nightmare
+          .wait('#input-agreement-search')
+          .insert('#input-agreement-search', agreement.name)
+          .click('[data-test-search-and-sort-submit]')
+          .wait(1000) // If another agreement was open wait for the new one to be open before the next operation.
```
doytch: Use `.waitUntilNetworkIdle(1000)` instead for things like this. (If you c/p'd it from my existing tests, oops, m'bad. I'm slowly fixing those :) )

test/ui-testing/tags-crud.js
```diff
@@ -0,0 +1,131 @@
...
+          .wait('#clickable-show-tags')
+          .click('#clickable-show-tags')
+          .wait('#input-tag-input')
+          .type('#input-tag-input', testTag)
+          .click('#input-tag-input')
+          .wait(2222)
+          .type('#input-tag-input', '\u000d')
+          .wait((tagValue) => {
+            return Array.from(
+              document.querySelectorAll('div[class*="valueChipValue"]')
```
doytch: This selector feels flaky given that we don't control the `class` being applied. Does the multiselect set `selectedValues`? Ie, could we use `querySelector('#input-tag-input').selectedValues` instead?

test/ui-testing/tags-crud.js
```diff
@@ -0,0 +1,131 @@
...
+            ).findIndex(e => e.textContent === tagValue) >= 0;
+          }, testTag)
+          .then(() => {
```
doytch: Why the nested `.then().catch()`?

test/ui-testing/tags-crud.js
```diff
@@ -0,0 +1,131 @@
...
+          })
+          .catch(done);
+      });
+
+      it('should have the badge count equal to 1', done => {
+        nightmare
+          .wait('#clickable-show-tags')
+          .wait(() => {
+            return Array.from(
+              document.querySelectorAll('span[class*="label"]')
```
doytch: Way too broad a selector here. Not only will this pass if any badge has a value of 1, it'll pass if any span with a class containing `label` (pretty common) _on the page_ has content of `1`. I think we can make this more specific.

test/ui-testing/tags-crud.js
```diff
@@ -0,0 +1,131 @@
...
+            ).findIndex(e => e.textContent === '1') >= 0;
+          })
+          .then(() => {
```
doytch: Same comment about nested then/catch.

src/components/Agreements/ViewAgreement/ViewAgreement.js
aditya-matukumalli: It basically reloads the _selectedAgreement_ resource. _path && path.match(/link/)_ checks if https://github.com/folio-org/stripes-erm-components/pull/40/files#diff-b846f11013fdd8210fb6bff3fe8b7e2cR24_ is getting mutated and if it is it returns _true_ triggers the refresh

src/components/Agreements/ViewAgreement/ViewAgreement.js
aditya-matukumalli: Good catch. Will change it

test/ui-testing/tags-crud.js
aditya-matukumalli: Doesn't look like. The **#input-tag-input** is only on the part of the text box thats empty and yet to be filled.

test/ui-testing/tags-crud.js
aditya-matukumalli: Yeah, right. Replacing it with
 _.evaluate(() => parseInt(document.querySelector('#clickable-show-tags').textContent) === 1)_

aditya-matukumalli:
@doytch Added in the requested changes

test/ui-testing/tags-crud.js
aditya-matukumalli: Added in the logic for incorporating further settings to that page by adding

```const list = document.querySelectorAll('[value="true"]');```

File: src/components/Agreements/ViewAgreement/ViewAgreement.js (hunk 1 above)
Comment (doytch) [reply src/components/Agreements/ViewAgreement/ViewAgreement.js]: Ahhh, I kept trying to figure out what _path_ had "link" in it and couldn't figure it out. It's the path _template_ that's getting matched. Good to know!

File: test/ui-testing/tags-crud.js (hunk 11 above)
Comment (doytch) [reply test/ui-testing/tags-crud.js]: C'est la vie!

File: test/ui-testing/tags-crud.js [hunk 15]
```diff
@@ -0,0 +1,98 @@
...
+          .click(config.select.settings)
+          .wait('#app-list-item-clickable-settings')
+          .wait('a[href="/settings/tags"]')
+          .click('a[href="/settings/tags"]')
+          .wait('a[href="/settings/tags/general"]')
+          .click('a[href="/settings/tags/general"]')
+          .wait('#tags_enabled')
+          .wait(222)
+          .evaluate(() => {
+            const list = document.querySelectorAll('[value="true"]');
```
doytch: Can we change this to `#tags_enabled[value="true"]`

aditya-matukumalli: Updated

PR #188, opened 2019-05-10
Erm 220

aditya-matukumalli:


src/components/Agreements/ViewAgreement/Sections/Organizations.js
```diff
@@ -30,14 +30,15 @@ export default class Organizations extends React.Component {
...
-              <Link to={`/vendors/view/${o.org.vendorsUuid}`}>{o.org.name}</Link> :
-              o.org.name
-            }
-            {o.role && `, ${o.role.label}`}
-          </Layout>
+        {orgs.map(o => (
+          o.org ?
+            <Layout className="marginTopHalf" key={o.org.id}>
+              {o.org.orgsUuid ?
+                <Link to={`/vendors/view/${o.org.orgsUuid}`}>{o.org.name}</Link> :
```
doytch: Needs to be updated.

aditya-matukumalli: Right. My bad.

src/components/Agreements/ViewAgreement/Sections/Organizations.js
```diff
@@ -30,14 +30,15 @@ export default class Organizations extends React.Component {
...
-          <Layout className="marginTopHalf" key={index}>
-            {o.org.vendorsUuid ?
-              <Link to={`/vendors/view/${o.org.vendorsUuid}`}>{o.org.name}</Link> :
-              o.org.name
-            }
-            {o.role && `, ${o.role.label}`}
-          </Layout>
+        {orgs.map(o => (
+          o.org ?
+            <Layout className="marginTopHalf" key={o.org.id}>
```
doytch: While you're in the code, could you tweak this `key` so that it can't be a duplicate? We allow an org to be assigned multiple times with different roles, so `org.id` isn't necessarily unique. `${o.org.id}-${o.role.value}` should be, though.

PR #216, opened 2019-06-14
Use the new Card component from STCOM to display interfaces, refs ERM…

aditya-matukumalli:
…-230


aditya-matukumalli:
Adding these changes as a part of a different PR

PR #228, opened 2019-07-03
 ERM-284: Manage 'Notes' on e-resources in Agreements

CalamityC:


test/ui-testing/package-notes.js
```diff
@@ -0,0 +1,15 @@
+const Pkg = require('./er-notes');
+
+const PKG = {
```
doytch: I'm not sure that hardcoding the package name here is the best way forward. The issue as I see it is that you're assuming that we'll always be running these tests on a platform that ran our particular start-up scripts which create that K-Int test package. 

That isn't a safe assumption - note that `folio-testing` doesn't even do that since [it doesn't have a K-Int Test Package.](https://folio-testing.aws.indexdata.com/erm/eresources?query=K-Int&sort=name).

I really like that you split off the testing code into `er-notes`, so I think it makes sense to just pass in a boolean like `isPackage`. Then, `er-notes` can check the **Is package: Yes** filter checkbox if `isPackage === true` and the **Is package: No** filter checkbox if `isPackage === false`. Then you can just pick the first result after the network goes idle and be assured that it's a package/title.

package.json
```diff
@@ -27,6 +27,8 @@
     "test-hide-eresources-func": "stripes test nightmare --run hide-eresources-func",
     "test-supplement": "stripes test nightmare --run supplement",
     "test-notes-crud": "stripes test nightmare --run notes-crud",
+    "test-package-notes": "stripes test nightmare --run package-notes",
```
doytch: Can you also add these two tests to the `test` and `test-int` `scripts` options just after the `notes-crud` test so that they run when we do a full test-suite run?

CalamityC:
I made all the changes and tried to add deleting the note type to er-notes test. But it's not always deleted. I noticed that the isPackage filter is not always properly working when Yes is checked (package-notes). As amends the note type is deleted at the end in that case. In the title-notes test it is not.

CalamityC:
Update: clicking reset all helps with the filter issue. 
But still no success with deleting note type in the title-notes test.

doytch:
So the issue with deleting the note _type_ it is that you can't because there's still a note using that type right? Because it wasn't possible to delete the note you created because the edit process failed, right? And that's because the edit process _appends_ to the name rather than replacing it?

So the appending appears to be the root problem, as far as I can tell. I've been trying to debug this and it's proving to be pretty ugly. I'm really not sure _why_ the note form doesn't respond to `.insert(..., '')` the way that all other text fields do. Additionally, there's some peculiarity about how it force-sets the title text field's value _a second time_, after it's set it once. I wonder if this is due to the Notes components using `react-final-form`... 

I may debug this at a later point, but for now, this is what I found worked:
```
      it(`should edit note ${testNote} to ${editedNote}`, done => {
        nightmare
          .evaluate(note => {
            const notesElements = [...document.querySelectorAll('div[role="gridcell"]')];
            const noteElement = notesElements.find(e => e.textContent === note);
            if (!noteElement) {
              throw Error(`Could not find row with the note ${note}`);
            }
            noteElement.click();
          }, testNote)
          .then(() => {
            nightmare
              .wait('[data-test-navigate-note-edit]')
              .click('[data-test-navigate-note-edit]')
              .wait(() => {
                const field = document.querySelector('[data-test-note-title-field]');
                return field && field.value ? true : false; // eslint-disable-line
              })
              // There's something odd happening with the way the NoteForm component fills
              // the title field. Even _after_ the field has shown a value (which we check for in
              // the above .wait()), the value is set _again_, which means we can't just clear the
              // TextField right away, we have to insert this ugly .wait(2000).
              .wait(2000)
              // Additionally, the TextField isn't responding to .insert(..., '') as a way of clearing
              // it, so we actually have to do it in JavaScript.
              .evaluate(() => { document.querySelector('[data-test-note-title-field]').value = ''; })
              .then(() => {
                nightmare
                  .insert('[data-test-note-title-field]', editedNote)
                  .wait('[data-test-save-note]')
                  .click('[data-test-save-note]')
                  .waitUntilNetworkIdle(2000)
                  .then(done);
              });
          })
          .catch(done);
      });```

That's really ugly though and I'm not a big fan of it. I'd rather we approach this differently all together. How about instead of attempting to clear the note title and enter a brand new one, we just edit the note title to be `${textNote}-edited`? That is, just append "-edited" to the name rather than trying to clear it. That way you can skip the two `.insert()`s and just `.type(... '-edited')` or something similar (perhaps using a variable?).

CalamityC:
> So the issue with deleting the note _type_ it is that you can't because there's still a note using that type right? Because it wasn't possible to delete the note you created because the edit process failed, right? And that's because the edit process _appends_ to the name rather than replacing it?

No, this is not the error I'm facing. The note with the note type is deleted and even the trash bin of the note type is clicked. But the last delete confirm button is not. And all that only happens when I test a title, with a package everything works.

BTW editing the note works for me, but I've noticed that the note type isn't always correctly selected when creating the note. But that doesn't matter for the note-type-delete-issue.

CalamityC:
It seems working after the latest commit, but I have no idea why. It's like voodoo.

PR #251, opened 2019-08-05
ERM-306: New and Edit Agreement panes: remove placeholder text

CalamityC:
remove empty option for agreement status dropdown
add empty placeholder for agreement status dropdown


src/components/AgreementSections/FormInfo.js
```diff
@@ -33,18 +33,13 @@ export default class FormInfo extends React.Component {
...
-      if (state.dataOptions[key].length !== (data[key].length + 1)) {
-        dataOptions[key] = [
-          { value: null, label: '' },
-          ...data[key],
-        ];
-      }
-    });
+    dataOptions.agreementStatusValues = [...data.agreementStatusValues];
+    dataOptions.renewalPriorityValues = [{ value: null, label: '' }, ...data.renewalPriorityValues];
+    dataOptions.isPerpetualValues = [{ value: null, label: '' }, ...data.isPerpetualValues];
```
doytch: This way of writing this function is kind of inefficient because it results in us creating brand new objects and arrays in state every single time this components gets props. We can rewrite this cleanly to avoid that by doing a few things:

1. Move the dataOptions into state as separate arrays. So our state would be initialized instead as:
```
state = {
  agreementStatusValues = [],
  renewalPriorityValues = [],
  isPerpetualValues = [],
}
```

2. In `gDSFP`, use the `newState` pattern. Eg:
```
static getDerivedStateFromProps(props, state) {
  const { data } = props;
  const newState = {};

  if (data.agreementStatusValues.length !== state.agreementStatusValues) {
    newState.agreementStatusValues = data.agreementStatusValues;
  }

  if (data.renewalPriorityValues.length + 1 !== state.renewalPriorityValues) {
    newState.renewalPriorityValues = [{ value: null, label: '' }, ...data.renewalPriorityValues];
  }

  if (data.isPerpetualValues.length + 1 !== state.isPerpetualValues) {
    newState.isPerpetualValues = [{ value: null, label: '' }, ...data.isPerpetualValues];
  }

  if (Object.keys(newState).length) return newState;

  return null;
}
```

3. Switch usages of the data options  in `render` from `state.dataOptions.xyzValues` to just `state.xyzValues`

CalamityC:
thanks for your comments, i made all changes

src/components/AgreementSections/FormInfo.js
```diff
@@ -24,35 +24,36 @@ export default class FormInfo extends React.Component {
...
   }
 
   // Prepend an empty value to each set of dropdown options to facilitate
   // unselecting a value.
   static getDerivedStateFromProps(props, state) {
     const { data } = props;
-    const dataOptions = {};
+    const newState = {};
+
+    if (data.agreementStatusValues.length !== state.agreementStatusValues) {
```
doytch: Sorry, these all should be `if (... != ....length)`. We're comparing whether the length of the array we just got from `props` has the same length (and thus is the same array) as the array we've stored in `state`.

CalamityC:
Sorry, I didn't see that. Fixed it.

PR #286, opened 2019-09-12
Move the Save & close button and add a Cancel button to Pane Footer, …

aditya-matukumalli:
…refs ERM-411, ERM-410

Moved the Save & close button and added a Cancel button to Pane footer


src/components/views/AgreementForm.js
```diff
@@ -128,6 +129,41 @@ class AgreementForm extends React.Component {
...
+        marginBottom0
+        onClick={this.props.handlers.onClose}
+      >
+        <FormattedMessage id="stripes-components.cancel" />
+      </Button>
+    );
+
+    const endButton = (
+      <Button
+        buttonStyle="primary paneHeaderNewButton"
```
doytch: Should this be `primary mega`? [Asking based on the ui-users example](https://github.com/folio-org/ui-users/blob/master/src/UserForm.js#L274).

src/components/views/AgreementForm.js
```diff
@@ -128,6 +129,41 @@ class AgreementForm extends React.Component {
...
+        disabled={this.props.pristine || this.props.submitting}
```
doytch: Should this be disabled if `props.invalid` as well?

PR #288, opened 2019-09-16
Preserve line-break characters for paragraph fields, refs ERM-444

aditya-matukumalli:
- Update stripes dependency to 2.10.1
- Update CHANGELOG


src/components/AgreementSections/LinkedLicenses.css
```diff
@@ -0,0 +1,3 @@
+.infoPopover {
+white-space: pre-wrap;
```
doytch:
```suggestion
  white-space: pre-wrap;
```

PR #289, opened 2019-09-18
Custom coverage tooltip for ERM-270

EthanFreestone:
Added tooltip functionality to the Custom Coverage Icon, commented out some seemingly useless code for review before deletion.


src/components/AgreementSections/LinesList.js
```diff
@@ -5,6 +5,7 @@ import { FormattedMessage } from 'react-intl';
 import { get } from 'lodash';
 import { MultiColumnList } from '@folio/stripes/components';
 import { Spinner } from '@folio/stripes-erm-components';
+import { Tooltip } from '@folio/stripes-components';
```
doytch: `Tooltip` should be imported via `@folio/stripes/components`. 

Anything that's re-exported by [`stripes`](https://github.com/folio-org/stripes) should never be imported directly, it should be pulled via Stripes itself. 

(I know this is kinda confusing when you see `stripes-erm-components` right above it, but `stripes-erm-components` isn't a part of Stripes [as you can see in Stripes' Github repo], but rather a package of ERM-related components for use in Stripes.)

src/components/AgreementSections/CoveredEResourcesList.js
```diff
@@ -61,7 +63,23 @@ export default class CoveredEResourcesList extends React.Component {
...
     accessEnd: () => 'TBD',
     coverage: e => <CoverageStatements statements={e.coverage} />,
-    isCustomCoverage: e => (e.customCoverage ? <CustomCoverageIcon /> : ''),
+    isCustomCoverage: line => {
+      const customCoverageTooltipLabel = 'Custom Coverage';
+      if (!line.customCoverage) return '';
+      return (
+        <Tooltip
+        text={customCoverageTooltipLabel}
+        id="custom_coverage_tooltip"
```
doytch: We prefer `kebab-case` ids over `snake_case`.

src/components/AgreementSections/LinesList.js
```diff
@@ -57,7 +60,23 @@ export default class LinesList extends React.Component {
...
     count: line => <EResourceCount resource={getResourceFromEntitlement(line)} />,
     coverage: line => <CoverageStatements statements={line.coverage} />,
-    isCustomCoverage: line => (line.customCoverage ? <CustomCoverageIcon /> : ''),
+    isCustomCoverage: line => {
+      const customCoverageTooltipLabel = 'Custom Coverage';
+      if (!line.customCoverage) return '';
+      return (
+          <Tooltip
+            text={customCoverageTooltipLabel}
+            id="custom_coverage_tooltip"
```
doytch: Same underscore/dash comment as above.

src/components/AgreementSections/LinesList.js
```diff
@@ -57,7 +60,23 @@ export default class LinesList extends React.Component {
     type: line => <EResourceType resource={getResourceFromEntitlement(line)} />,
     count: line => <EResourceCount resource={getResourceFromEntitlement(line)} />,
     coverage: line => <CoverageStatements statements={line.coverage} />,
-    isCustomCoverage: line => (line.customCoverage ? <CustomCoverageIcon /> : ''),
+    isCustomCoverage: line => {
+      const customCoverageTooltipLabel = 'Custom Coverage';
```
doytch: This needs to be an i18n'd string via `FormattedMessage`.

src/components/AgreementSections/CoveredEResourcesList.js
```diff
@@ -61,7 +63,23 @@ export default class CoveredEResourcesList extends React.Component {
     accessStart: () => 'TBD',
     accessEnd: () => 'TBD',
     coverage: e => <CoverageStatements statements={e.coverage} />,
-    isCustomCoverage: e => (e.customCoverage ? <CustomCoverageIcon /> : ''),
+    isCustomCoverage: line => {
+      const customCoverageTooltipLabel = 'Custom Coverage';
```
doytch: This needs to be an i18n'd string via `FormattedMessage`.

src/components/CustomCoverageIcon/CustomCoverageIcon.js
```diff
@@ -1,16 +1,17 @@
...
-        >
+
+const CustomCoverage = React.forwardRef((props, ref) => (
+  <FormattedMessage id="ui-agreements.agreementLines.hasCustomCoverage">
+    {() => (
+      <Layout
+        className="flex"
+        data-test-custom-coverage
+      >
+        <button type="button" ref={ref} aria-labelledby={props['aria-labelledby']}>
```
doytch: This shouldn't be a `button`. Perhaps a `span` instead.

A button is a very specific thing that implies certain behaviour: it's clickable and does something when it's clicked. Custom coverage icons don't do either of those things so calling a button is bad form since it can be confusing to screen readers and can throw off things like CSS rules/selectors.

translations/ui-agreements/en_GB.json
```diff
@@ -220,5 +220,6 @@
     "license.unlinkLicense": "Unlink license",
     "license.noLicenses": "Agreement has no licenses.",
     "agreements.internalContactsRole": "Internal contact role",
-    "eresources.defaultCoverage": "Default coverage"
+    "eresources.defaultCoverage": "Default coverage",
+    "customcoverages.tooltip": "Custom Coverage"
```
doytch: Setting it doesn't make a real difference, but you don't need to specify translation keys for languages other than `en`. They'll eventually be run through and generated.

src/components/AgreementSections/LinesList.js
```diff
@@ -14,6 +14,7 @@ import EResourceProvider from '../EResourceProvider';
 import EResourceType from '../EResourceType';
 import { getResourceFromEntitlement, urls } from '../utilities';
 
+
```
doytch: Be careful about extra blank space.

src/components/AgreementSections/LinesList.js
```diff
@@ -39,6 +40,7 @@ export default class LinesList extends React.Component {
   }
 
   formatter = {
+
```
doytch: Ditto above.

src/components/AgreementSections/LinesList.js
```diff
@@ -57,7 +59,18 @@ export default class LinesList extends React.Component {
     type: line => <EResourceType resource={getResourceFromEntitlement(line)} />,
     count: line => <EResourceCount resource={getResourceFromEntitlement(line)} />,
     coverage: line => <CoverageStatements statements={line.coverage} />,
-    isCustomCoverage: line => (line.customCoverage ? <CustomCoverageIcon /> : ''),
+    isCustomCoverage: line => {
+      if (!line.customCoverage) return '';
+      return (
+        <Tooltip
+          text={<FormattedMessage id="ui-agreements.customcoverages.tooltip" />}
```
doytch: We prefer to alphasort our props in ERM modules. Ie, `id` before `text`. Just makes it easier to find props when dealing with components that have a bunch of them.

src/components/AgreementSections/CoveredEResourcesList.js
```diff
@@ -19,6 +20,7 @@ import EResourceLink from '../EResourceLink';
 import FormattedUTCDate from '../FormattedUTCDate';
 import IfEResourcesEnabled from '../IfEResourcesEnabled';
 
+
```
doytch: Avoid multiple blank lines.

src/components/AgreementSections/CoveredEResourcesList.js
```diff
@@ -58,9 +60,20 @@ export default class CoveredEResourcesList extends React.Component {
...
     package: e => get(e._object, 'pkg.name', '-'),
     coverage: e => <CoverageStatements statements={e.coverage} />,
-    isCustomCoverage: e => (e.customCoverage ? <CustomCoverageIcon /> : ''),
     accessStart: e => this.renderDate(get(e._object, 'accessStart')),
     accessEnd: e => this.renderDate(get(e._object, 'accessEnd')),
+    isCustomCoverage: line => {
+      if (!line.customCoverage) return '';
+      return (
+        <Tooltip
+          text={<FormattedMessage id="ui-agreements.customcoverages.tooltip" />}
```
doytch: We prefer to alphasort our props in ERM modules. Ie, `id` before `text`. Just makes it easier to find props when dealing with components that have a bunch of them.

src/components/CustomCoverageIcon/CustomCoverageIcon.js
EthanFreestone: Ah, this was on the suggestion of Rasmus, who mentioned that tooltips need to be contained inside focusable elements, is there something I should use instead of span in this case, or is there something I can do to span to make it focusable?

translations/ui-agreements/en_GB.json
EthanFreestone: Ok, should I leave those for now anyway, or go through and remove?

src/components/CustomCoverageIcon/CustomCoverageIcon.js
doytch: Ahhh, makes sense. Personally I'd try setting `tabindex="-1"` on the span first. That makes the element focusable via programmatic means (which I assume is what's necessary for the Tooltip), but will not be tabbed through via the keyboard.

If that fails, then we can just go ahead and just use `button`, but add a comment above it to explain why this is in fact a button.

translations/ui-agreements/en_GB.json
doytch: Nah, leave em. Just trying to save you time for the future.

src/components/CustomCoverageIcon/CustomCoverageIcon.js
EthanFreestone: That appears to have worked, closing this thread

src/components/AgreementSections/CoveredEResourcesList.js
```diff
@@ -58,9 +59,20 @@ export default class CoveredEResourcesList extends React.Component {
...
     package: e => get(e._object, 'pkg.name', '-'),
     coverage: e => <CoverageStatements statements={e.coverage} />,
-    isCustomCoverage: e => (e.customCoverage ? <CustomCoverageIcon /> : ''),
     accessStart: e => this.renderDate(get(e._object, 'accessStart')),
     accessEnd: e => this.renderDate(get(e._object, 'accessEnd')),
+    isCustomCoverage: line => {
+      if (!line.customCoverage) return '';
+      return (
+        <Tooltip
+          id="custom-coverage-tooltip"
```
doytch: Just realised that this `id` won't be unique across the page while it should. So maybe something like this:
```suggestion
          id={`covered-eresources-cc-tooltip-${line.rowIndex}`}
```

(`rowIndex` is provided by `MultiColumnList`)

src/components/AgreementSections/LinesList.js
```diff
@@ -57,7 +57,18 @@ export default class LinesList extends React.Component {
     type: line => <EResourceType resource={getResourceFromEntitlement(line)} />,
     count: line => <EResourceCount resource={getResourceFromEntitlement(line)} />,
     coverage: line => <CoverageStatements statements={line.coverage} />,
-    isCustomCoverage: line => (line.customCoverage ? <CustomCoverageIcon /> : ''),
+    isCustomCoverage: line => {
+      if (!line.customCoverage) return '';
+      return (
+        <Tooltip
+          id="custom-coverage-tooltip"
```
doytch: Just realised that this `id` won't be unique across the page while it should. So maybe something like this:
```suggestion
          id={`agreement-line-cc-tooltip-${line.rowIndex}`}
```

(`rowIndex` is provided by `MultiColumnList`)

dltj:
Please do not make changes to translation files other than `en.json`.  They will be overwritten (and in this case caused a conflict with the automated lokalise-to-GitHub script).

EthanFreestone:
Apologies, Mark had mentioned that. Shall bear in mind in future.

PR #292, opened 2019-09-19
ERM-428 ERM-429: Manage and view linked license amendments

doytch:
- Moved linked licenses into separate Controlling, Historical and Future Licenses accordions.
- Moved external licenses into its own accordion.
- Created `LinkedLicenseCard` component to render the linked license as a card that contains the license and amendment info.
- Created `getLicenseAmendments` utility to split out amendments for the various renderers.
- Created `LicenseAmendmentList` component to render....license amendments as a list.
- Created `AmendmentsFieldArray` component to enable configuration of the amendment status and note.
  - Various validators and warn functions.
  - Updated `LicensesFieldArray` to use `AmendmentsFieldArray`

Tests
- Fixed a bunch of tests that broke with the switch to `Headline` for agreement/license name.
- Fixed org interface tests.
- Updated and augmented linked license tests to handle the new accordions and amendment lists.

Misc
- Converted a bunch of hardcoded URLs to be fetched from the URL generators.
- Added accessStart and accessEnd columns for covered eresources (ERM-395)


src/components/views/AgreementForm.js
```diff
@@ -147,10 +147,10 @@ class AgreementForm extends React.Component {
     const endButton = (
       <Button
         buttonStyle="primary mega"
-        disabled={this.props.pristine || this.props.submitting || this.props.invalid}
+        disabled={pristine || submitting}
```
aditya-matukumalli: May want to add _invalid_ as well? Then when you render an empty card on clicking the _Add license to agreement_ button, the _save and close_ button would still be disabled.

doytch: So I actually remove `invalid` on purpose...

If you link a license, then you have to set that license's status and the status of its amendments. However, `Select` doesn't show the `error` it gets until the field is `touched`. So if you just add a license, then the Save button is disabled and you have no idea why unless you notice the little required asterisk.

Now that I've enabled the button if the form is invalid, when the user clicks Save, the errors will get shown and the form will scroll to the errors themselves. This is much more friendly to the user. I think we should consider making this change in ui-licenses and ui-local-kb-admin as well...

That's not really strong UX IMO, so

aditya-matukumalli: Ah okay. That makes sense. Cool.

PR #293, opened 2019-09-24
ERM-456: Add activeFrom and activeTo datepickers on Agreement lines

aditya-matukumalli:
- Manage Agreement line Active dates
- Tests


src/components/AgreementLinesFieldArray/AgreementLineField.js
```diff
@@ -123,6 +150,32 @@ export default class AgreementLineField extends React.Component {
             </KeyValue>
           </Col>
         </Row>
+        <Row>
+          <Col xs={12} md={4}>
+            <Field
+              backendDateStandard="YYYY-MM-DD"
+              component={Datepicker}
+              dateFormat="YYYY-MM-DD"
+              id="agreement-line-active-from"
```
doytch: This id should be unique, perhaps containing the `index`.

src/components/AgreementLinesFieldArray/AgreementLineField.js
```diff
@@ -123,6 +150,32 @@ export default class AgreementLineField extends React.Component {
...
+              name={`${name}.activeFrom`}
+              validate={this.validateDateOrder}
+            />
+          </Col>
+          <Col xs={12} md={4}>
+            <Field
+              backendDateStandard="YYYY-MM-DD"
+              component={Datepicker}
+              dateFormat="YYYY-MM-DD"
+              id="agreement-line-active-to"
```
doytch: This id should be unique, perhaps containing the `index`.

src/components/AgreementLinesFieldArray/AgreementLineField.js
```diff
@@ -123,6 +150,32 @@ export default class AgreementLineField extends React.Component {
...
           </Col>
         </Row>
+        <Row>
+          <Col xs={12} md={4}>
+            <Field
+              backendDateStandard="YYYY-MM-DD"
+              component={Datepicker}
+              dateFormat="YYYY-MM-DD"
+              id="agreement-line-active-from"
+              index={index}
```
doytch: No need to pass an `index` to this, is there?

src/components/AgreementLinesFieldArray/AgreementLineField.js
```diff
@@ -123,6 +150,32 @@ export default class AgreementLineField extends React.Component {
...
+              validate={this.validateDateOrder}
+            />
+          </Col>
+          <Col xs={12} md={4}>
+            <Field
+              backendDateStandard="YYYY-MM-DD"
+              component={Datepicker}
+              dateFormat="YYYY-MM-DD"
+              id="agreement-line-active-to"
+              index={index}
```
doytch: No need to pass an `index` to this, is there?

src/components/AgreementLinesFieldArray/AgreementLineField.js
```diff
@@ -38,6 +38,33 @@ export default class AgreementLineField extends React.Component {
...
+      } else if (name.indexOf('endDate') >= 0) {
+        activeFrom = new Date(get(allValues, name.replace('activeTo', 'activeFrom')));
+        activeTo = new Date(value);
+      } else {
+        return undefined;
+      }
+
+      if (activeFrom >= activeTo) {
+        return (
+          <div data-test-error-end-date-too-early>
```
doytch: Maybe throw in a test case to validate that this is working?

aditya-matukumalli: I think test for this would look out of place here, as we are depending on the PO lines tests here to test this functionality. A better idea would be to enforce this validation on the start and end dates while creating an agreement, and writing a test for it? That way we would have this validation logic covered in both agreements and licenses.

doytch: I don't understand what PO lines have to do with this? I don't really care where the test is, but yeah I agree with you it makes sense to put it in the `agreement-crud` (or maybe actually the `basket`) test.

(In case it's confusing, I only put this comment down on this line just because I saw there were `data-test-` attributes that we could use to test if this validator is working correctly.)

aditya-matukumalli: Yeah, I meant to say I didn't add the test in the PO lines tests, but that test functionality would run only when the PO lines tests run, as we are calling the _createAgreement_ function from there with a basket id passed to it, where then we populate the date fields.

aditya-matukumalli: Add the validation test from within PO line as it was the easiest place to put in for this specific
functionality this PR covers.

src/components/AgreementSections/LinesList.js
```diff
@@ -84,9 +84,8 @@ export default class LinesList extends React.Component {
 
   renderPOLine = (line) => {
     const { orderLines } = this.props.agreement;
-
     if (!line.poLineId) return '';
-    if (!orderLines) return <Spinner />;
+    if (!orderLines.length) return <Spinner />;
```
doytch: `orderLines` can be undefined based on `AgreementViewRoute:190`. It specifically _doesn't_ have a default `[]` value so that we can snoop whether the fetch has returned yet.

So I think this can crash, can't it?

PR #300, opened 2019-09-30
ERM-466 ERM-467 ERM-477: Manage and view agreement periods

doytch:
Agreement start/end/cancellation dates are no longer set on the agreement, they're set on a `period` which is an entry in an agreement's `periods` array.

- Added new `AgreementPeriodsFieldArray` component to enable managing the periods on the forms.
- Tweaked existing start/end/cancellation labels to show that they're referring to the current period.
  - The current period is calculated by the backend and the existing start/end/cancellation fields are populated with the current period's values for those fields.
- Added new **Other periods** accordion to show periods that aren't the current period.
- Updated tests.
- Added support for erm interface 2.0 which changes the permissions sets: https://github.com/folio-org/mod-agreements/pull/185


doytch:
Parking this with `do-not-merge` since it requires the backend to be merged (with it's breaking changes and new interface version).

PR #301, opened 2019-10-01
ERM-396: Viewing upcoming joiners/leavers for a package

CalamityC:


src/routes/AgreementViewRoute.js
```diff
@@ -220,6 +222,11 @@ class AgreementViewRoute extends React.Component {
     this.props.history.push(`${urls.agreements()}${this.props.location.search}`);
   }
 
+  handleClickFilterButton = (path) => {
```
doytch: I think `handleFilterEResources` might be more explanatory of what we're doing with this filter. An agreement is so big so it's helpful to have pretty specific names for our callbacks/props.

src/routes/AgreementViewRoute.js
```diff
@@ -44,6 +44,7 @@ class AgreementViewRoute extends React.Component {
       records: 'results',
       recordsRequired: '%{agreementEresourcesCount}',
     },
+    filterPath: { initialValue: 'current' },
```
doytch: `eresourcesFilterPath` might be more self-explanatory and less likely to clash with future filters we need to implement.

src/routes/AgreementViewRoute.js
```diff
@@ -315,10 +322,12 @@ class AgreementViewRoute extends React.Component {
         canEdit={this.props.stripes.hasPerm('ui-agreements.agreements.edit')}
         data={{
           agreement: this.getCompositeAgreement(),
+          filterPath: this.props.resources.filterPath,
           terms: get(resources, 'terms.records', []),
         }}
         handlers={{
           ...handlers,
+          onClickFilterButton: this.handleClickFilterButton,
```
doytch: So if we did the renaming above, this would end up being `onFilterEResources`

src/routes/EResourceViewRoute.js
```diff
@@ -45,6 +45,7 @@ class EResourceViewRoute extends React.Component {
     },
     query: {},
     entitlementsCount: { initialValue: RECORDS_PER_REQUEST },
+    filterPath: { initialValue: 'current' },
```
doytch: Same sort of comments as the variables in `AgreementViewRoute`. Maybe `packageContentsFilter`, `handleFilterPackageContents` and `onFilterPackageContents`.

src/components/EResourceSections/PackageContents.js
```diff
@@ -3,35 +3,66 @@ import PropTypes from 'prop-types';
...
+    name: <FormattedMessage id="ui-agreements.eresources.name" />,
+    platform: <FormattedMessage id="ui-agreements.eresources.platform" />,
+    coverage: <FormattedMessage id="ui-agreements.eresources.coverage" />,
+    accessStart: <FormattedMessage id="ui-agreements.eresources.accessStart" />,
+    accessEnd: <FormattedMessage id="ui-agreements.eresources.accessEnd" />,
+  }
+
+  columnWidths = {
+    name: 250,
+    platform: 150,
```
doytch: Should we define a width for `platform`? My default these days is to leave widths undefined because MCL is much better at calculating widths automatically, but if you've seen it consistently sizing it poorly, this makes sense.

src/components/EResourceSections/PackageContents.js
```diff
@@ -3,35 +3,66 @@ import PropTypes from 'prop-types';
...
+  ]
+
+  renderList = (packageContents) => {
+    return (
+      <MultiColumnList
+        columnMapping={this.columnMapping}
+        columnWidths={this.columnWidths}
+        contentData={packageContents}
+        formatter={this.formatter}
+        id="packageContents-list"
```
doytch: Prefer kebab-casing eg, `package-contents-list`

src/components/AgreementSections/CoveredEResourcesList.js
```diff
@@ -51,7 +55,7 @@ export default class CoveredEResourcesList extends React.Component {
     package: 150,
   }
 
-  formatter = {
+  formatter= {
```
doytch: Nitpick...keep spaces around the `=` 😆

PR #307, opened 2019-10-08
PR for FrontEnd of ERM-465

EthanFreestone:


src/components/AgreementSections/FormInfo.js
```diff
@@ -15,25 +15,34 @@ import {
 import AgreementPeriodsFieldArray from '../AgreementPeriodsFieldArray';
 import { validators } from '../utilities';
 
-export default class FormInfo extends React.Component {
+class FormInfo extends React.Component {
```
doytch: Keep the `export default` with the `class` declaration if there's no particular to separate it (like you would need to do if you were wrapping the class in a higher-order component). 

Whenever we see a `class` lacking the `export default` it signals that it's gonna be wrapped later, which we're not doing here.

src/components/AgreementSections/FormInfo.js
```diff
@@ -48,75 +57,123 @@ export default class FormInfo extends React.Component {
       newState.isPerpetualValues = [{ value: null, label: '' }, ...data.isPerpetualValues];
     }
 
+    if (data.reasonForClosureValues.length + 1 !== state.reasonForClosureValues.length) {
+      newState.reasonForClosureValues = [{ value: null, label: '' }, ...data.reasonForClosureValues];
+    }
+
+    // Check if status is closed and, if so, change state to allow reasonForClosure field
+    if (values.agreementStatus === 'closed') {
```
doytch: Can you pop an entry for `closed` into `ui-agreements/src/constants/statuses.js` and use that?

src/components/AgreementSections/Info.js
```diff
@@ -28,7 +28,7 @@ export default class Info extends React.Component {
 
   render() {
     const { agreement } = this.props;
-
+    const agreementIsClosed = (agreement.agreementStatus !== null && (agreement.agreementStatus.value === 'closed'));
```
doytch: Same thing as above, an entry for `closed` in `ui-agreements/src/constants/statuses.js` would be nicer.

doytch: Also, you can make this a bit more succinct as `const agreementIsClosed = get(agreement, 'agreementStatus.value') === statuses.CLOSED;`

doytch: Bonus, that also saves you from the inevitable crash when `agreementStatus` is merely `undefined` and not `null`.

src/components/views/AgreementForm.js
```diff
@@ -217,6 +219,9 @@ class AgreementForm extends React.Component {
 export default stripesFinalForm({
   initialValuesEqual: (a, b) => isEqual(a, b),
   keepDirtyOnReinitialize: true,
+  subscription: {
```
doytch: @aditya-matukumalli, is this necessary if he's using a `FormSpy`. It feels like we're subscribing to the same thing (`values`) twice. If that's the case, it'd be nice if we removed one of them for optimization.

aditya-matukumalli:
@doytch Yeah, looks like we can get away with just once by subscribing at the top level of the form and then passing `values` down to the `onChange` on `FormSpy` instead of subscribing from within `FormSpy`

aditya-matukumalli:
@doytch Looks like https://final-form.org/docs/react-final-form/types/FormSpyProps#onchange from here, for the `onChange` to fire it would require a `subscription`, but that does not give us a `Form` level subscription which we are explicitly doing at the form level. It maybe possible but I'd have to dig in more. Will do this tomorrow.

src/components/AgreementSections/FormInfo.js
```diff
@@ -48,75 +57,123 @@ export default class FormInfo extends React.Component {
...
-          </Col>
-        </Row>
-        <FieldArray
-          component={AgreementPeriodsFieldArray}
-          name="periods"
-        />
-      </div>
-    );
-  }
+    warnReason = ({ values }) => {
```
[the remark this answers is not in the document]
EthanFreestone: Cleaned up the indenting

src/components/AgreementSections/FormInfo.js
```diff
@@ -48,75 +57,123 @@ export default class FormInfo extends React.Component {
...
+      const { form } = this.props;
+
+      if (values.reasonForClosure && values.agreementStatus !== 'closed') {
```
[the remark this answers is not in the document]
EthanFreestone: I have done this now, I was wondering though how this knows to set our value 'warning' to the intrinsic warning property of Field?

src/components/AgreementSections/FormInfo.js
EthanFreestone: Ah right, I did this to resolve the lint error, but I can see that a newline would have sufficed. Done now.

src/components/AgreementSections/FormInfo.js
EthanFreestone: This has now been done.

src/components/AgreementSections/Info.js
EthanFreestone: Done as part of same commit as above

EthanFreestone: Thank you, I have done so now

src/components/AgreementSections/FormInfo.js
```diff
@@ -48,14 +58,39 @@ export default class FormInfo extends React.Component {
...
+    let warning;
+
+    if (values.reasonForClosure && values.agreementStatus !== statuses.CLOSED) {
+      warning = (
+        <div data-test-warn-clear-reason-for-closure>
+          <FormattedMessage id="ui-agreements.warn.clearReasonForClosure" />
+        </div>
+      );
+    } else {
+      warning = undefined;
```
doytch: This `else` case isn't necessary. If you `let foo;`, then `foo` is `undefined` from the start.

src/components/AgreementSections/FormInfo.js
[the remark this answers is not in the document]
doytch: `{ foo }` is just JS shorthand for fully writing out `{ foo: foo }`.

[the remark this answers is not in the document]
doytch: [The ECMAScript 2015 section of these docs shows some examples.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#Syntax)

doytch:
Pushed a commit that crucially:
- Adds reloading of the saved `reasonForClosure` value. Previously, if you saved a value and then hit Edit again, the saved value wouldn't be autoselected on the form.
- Adds the ability to clear the selected value. Had to add in an extra `parse` since RFF coalesces empty strings to `undefined` by default. We don't want that, we want to send an empty string to the backend so it knows to unset the value.
- Removed the `state` usage since it was unnecessary.

PR #308, opened 2019-10-09
Display earliest agreement period, refs ERM-497

aditya-matukumalli:


src/components/AgreementSections/OtherPeriods.js
```diff
@@ -46,13 +44,14 @@ export default class Lines extends React.Component {
   render() {
     const {
       agreement,
+      agreement: { currentPeriod },
```
doytch: I suppose it'd be marginally cleaner if just wrote the following and then we could cut out both of the `||` and `&&` checks on lines 53 and 54.
```suggestion
      agreement: { currentPeriod = {}, periods = [] },
```

PR #311, opened 2019-10-10
Display grey icon where agreement status is closed, refs ERM-484

aditya-matukumalli:


src/components/views/Agreements.js
```diff
@@ -89,6 +89,15 @@ export default class Agreements extends React.Component {
   }
 
   formatter = {
+    name: a => (
+      <AppIcon
+        app="agreements"
+        size="small"
+        className={get(a, 'agreementStatus.label') !== 'Closed' ? '' : css.inactiveAppIcon}
```
doytch: Can you have this key off of the `value` and check it [against the constants](https://github.com/folio-org/ui-agreements/blob/master/src/constants/statuses.js#L2)?

PR #317, opened 2019-10-14
ERM-501: Fixed error messages for date fields

doytch:
Generalising the validators also resulted in reusing the same error messages...whoops! This keeps the logic general but allows the message to be specified external to the method.


doytch:
@dltj, is it safe to fix the string in `en_US` here? Otherwise the changes won't be visible during usage.

dltj:
We'll effectively have a no-op.  The changes to `en_US.json` would get overwritten in the next Lokalise run, but since what will come from Lokalise is what is in the new `en.json` it would be an empty diff.

PR #331, opened 2019-10-23
Erm 503

aditya-matukumalli:


src/components/AgreementSections/CoveredEResourcesList.js
```diff
@@ -61,6 +63,11 @@ export default class CoveredEResourcesList extends React.Component {
       const titleInstance = get(e._object, 'pti.titleInstance', {});
       return <EResourceLink eresource={titleInstance} />;
     },
+    issn: e => {
+      const titleInstance = get(e._object, 'pti.titleInstance', {});
+      const eissn = getResourceIdentifier(titleInstance, 'eissn');
```
doytch: If you wanted to be slicker, you could always write this as `return getResourceIdentifier(titleInstance, 'eissn') || gRI(tI, 'pissn');`

PR #339, opened 2019-10-29
BigTests infrastructure setup in agreements

aditya-matukumalli:


test/bigtest/tests/package-content-test.js
```diff
@@ -0,0 +1,110 @@
...
+  name: 'access_start_access_end_tests Package'
+};
+
+const testData = [
+  { pti: { titleInstance: { id: '1234', name: 'testInstance1' } }, accessStart: () => faker.date.recent().toISOString() },
+  { pti: { titleInstance: { id: '234', name: 'testInstance2' } }, accessStart: () => faker.date.future().toISOString() },
+  { pti: { titleInstance: { id: '734', name: 'testInstance3' } }, accessEnd: () => faker.date.recent().toISOString() },
+];
+
+describe('PackageContentCRUD', () => {
```
doytch: Can you pick a better name for this? You're not really testing CRUD operations here are you?

test/bigtest/tests/package-content-test.js
```diff
@@ -0,0 +1,110 @@
...
+
+import setupApplication from '../helpers/setup-application';
+import PackageContentInteractor from '../interactors/package-content';
+
+const PKG = {
+  name: 'access_start_access_end_tests Package'
+};
+
+const testData = [
+  { pti: { titleInstance: { id: '1234', name: 'testInstance1' } }, accessStart: () => faker.date.recent().toISOString() },
```
doytch: A bit clearer naming here will make maintaining these tests easier. We can do better than `testInstance#`

test/bigtest/network/factories/eresource.js
```diff
@@ -0,0 +1,21 @@
+import { Factory, faker } from '@bigtest/mirage';
+
+export default Factory.extend({
+  id: () => faker.random.uuid(),
+  class: () => faker.random.arrayElement(['org.olf.kb.Pkg']),
+  name: () => faker.random.words(),
+  customCoverage: () => false,
+
+
+  afterCreate(eresource, server) {
```
doytch: This (and the `class` above, I guess), seems to suggest that all eresources are packages. I know currently they are, but we know they won't be for all tests so we're doing ourselves a disservice by not thinking about that future scenario.

test/bigtest/network/factories/eresource.js
```diff
@@ -0,0 +1,21 @@
...
+    eresource.testData.forEach(item => {
```
doytch: Maybe given the above, we can also change this from `testData`? What's `testData` mean? Normally I'd assume some very generic thing, but here it's literally an array of PCIs. Simply by calling it something else, we'd be better accomplishing the segmentation we're going to need to handle non-package eresources.

test/bigtest/tests/package-content-test.js
```diff
@@ -0,0 +1,110 @@
...
+      this.visit(`/erm/eresources/${eresource.id}`);
+    });
+
+    describe('visiting the eresource pane', () => {
+      it('should render expected package', () => {
+        expect(packagecontent.headline).to.equal(PKG.name);
+      });
+
+      it('should render the expected current eresource', () => {
+        expect(packagecontent.eresourceName).to.equal(testData[0].pti.titleInstance.name);
```
doytch: Avoid magic numbers like these indices into `testData` you use throughout. It makes it hard to read the first time and it'll be unimaginably harder for someone to parse through a few years from now.

We could write these sections this instead as:
```
const droppedResource = ...
const currentResource = ...
const futureResource = ...
...
package = this.server.create('eresource', { name: packageName, pcis: [ droppedResource, currentResource, futureResource ] });
...
expect(packageContent.eresourceName).to.equal(futureResource.pti.titleInstance.name)
```

test/bigtest/interactors/package-content.js
```diff
@@ -0,0 +1,37 @@
+import {
+  interactor,
+  clickable,
+  fillable,
+  isPresent,
+  selectable,
+  text,
+} from '@bigtest/interactor';
+
+export default @interactor class PackageContentInteractor {
```
doytch: This interactor is waaaaaay overloaded with responsibilities. This Package Content interactor has methods related to, yes, the package contents, but also managing the Basket, filling out Agreement forms, and interacting with an Agreement's view?

package.json
```diff
@@ -14,11 +14,14 @@
     "start": "stripes serve",
     "build": "stripes build --output ./output",
 
-    "test": "stripes test nightmare --run nav/agreement-crud/contacts/orgs/interfaces/basket/tags-crud/external-licenses/custom-coverage/hide-eresources-func/supplement/agreement-notes/package-notes/title-notes/related-agreements",
-    "test-erm": "stripes test nightmare --run nav/agreement-crud/contacts/orgs/interfaces/basket/tags-crud/external-licenses/custom-coverage/hide-eresources-func/supplement/agreement-notes/package-notes/title-notes/related-agreements/linked-licenses/license-terms",
+    "test": "stripes test karma --karma.singleRun; stripes test nightmare --run nav/agreement-crud/contacts/orgs/interfaces/basket/tags-crud/external-licenses/custom-coverage/hide-eresources-func/supplement/notes-crud/package-notes/title-notes/related-agreements",
```
doytch:
```suggestion
    "test": "yarn run test-nightmare; yarn run test-bigtest-once",
```

package.json
```diff
@@ -14,11 +14,14 @@
...
+    "test-nightmare": "stripes test nightmare --run nav/agreement-crud/contacts/orgs/interfaces/basket/tags-crud/external-licenses/custom-coverage/hide-eresources-func/supplement/notes-crud/package-notes/title-notes/related-agreements",
+    "test-bigtest": "stripes test karma",
+    "test-bigtest-once": "stripes test karma --karma.singleRun",
+    "test-erm": "stripes test nightmare --run nav/agreement-crud/contacts/orgs/interfaces/basket/tags-crud/external-licenses/custom-coverage/hide-eresources-func/supplement/notes-crud/package-notes/title-notes/related-agreements/linked-licenses/license-terms",
```
doytch:
```suggestion
    "test-erm": "yarn run test-nightmare; yarn run test-licenses-int",
```

test/bigtest/interactors/package-content.js
doytch: There's still too much stuff in here. We're still having a `PackageContentInteractor` responsible for interacting with things that aren't even in its _route_ (`/erm/eresources` vs `/erm/eresources/{pkgId}`).

Honestly, it looks like the only things here that should exist in this interactor are the `click***` actions. They're the only things here that are truly part of the `PackageContents` component.

doytch: While you're at it, you may as well add the `clickAll` and `clickCurrent` handlers though 🤷‍♂

test/bigtest/tests/package-content-test.js
```diff
@@ -0,0 +1,126 @@
...
+      });
+
+      describe('Add eresource to basket and create agreement', () => {
+        beforeEach(async function () {
+          await basket.addToBasketButtion();
+          await basket.clickOpenBasket();
+          await basket.clickCreateNewAgreement();
+          await agreementform.fillName('testAgreement');
+          await agreementform.selectStatus('Draft');
+          await agreementform.processDate('2020-04-01');
```
doytch: Just looking at this, as someone who doesn't _really_ know what rules agreement form's have, I'd have no idea what `processDate()` does. Consistency is important, can you call it something like `fillStartPeriod`? Should it also be referring to a specific period (it's hardcoded to `0` at the moment)?

test/bigtest/tests/package-content-test.js
```diff
@@ -0,0 +1,126 @@
...
+        expect(packagecontent.eresourceName).to.equal(currentResource.pti.titleInstance.name);
+      });
+
+      describe('clicking the future eresource tab', () => {
+        beforeEach(async function () {
+          await packagecontent.clickFuture();
+        });
+
+        it('should render the expected future resource', () => {
+          expect(packagecontent.eresourceName).to.equal(futureResource.pti.titleInstance.name);
```
doytch: There's a big mix of all lowercase and camelCase in these tests. Is there a style reason why you tend to lowercase the interactor variable names?

test/bigtest/tests/package-content-test.js
```diff
@@ -0,0 +1,126 @@
...
+          await basket.clickCreateNewAgreement();
+          await agreementform.fillName('testAgreement');
+          await agreementform.selectStatus('Draft');
+          await agreementform.processDate('2020-04-01');
+          await agreementform.createAgreement();
+          await agreementview.whenLoaded();
+        });
+
+        describe('should open agreement view pane', () => {
+          describe('clicking current eresource', () => {
```
doytch: You're not `clicking` the current eresource. Again, _I_ know what you mean, but this needs to be easily understandable in the future by new developers.

test/bigtest/tests/package-content-test.js
```diff
@@ -0,0 +1,126 @@
...
+          await basket.clickOpenBasket();
+          await basket.clickCreateNewAgreement();
+          await agreementform.fillName('testAgreement');
+          await agreementform.selectStatus('Draft');
+          await agreementform.processDate('2020-04-01');
+          await agreementform.createAgreement();
+          await agreementview.whenLoaded();
+        });
+
+        describe('should open agreement view pane', () => {
```
doytch: Do we reserve the `should` "keyword" for `it` cases?

test/bigtest/tests/package-content-test.js
```diff
@@ -0,0 +1,126 @@
...
+
+            it('should render the expected current eresource', () => {
+              expect(packagecontent.eresourceName).to.equal(currentResource.pti.titleInstance.name);
+            });
+          });
+
+          describe('clicking future eresource', () => {
+            beforeEach(async function () {
+              await agreementview.clickLinesAccordion();
+              await packagecontent.clickFuture();
```
doytch: In an ideal world, this would probably be something like `await agreementView.linesSection.coveredEresourcesList.clickFuture()` but it's up to you. 

Because of course, we're _not_ on the package content view, and it's just good fortune that the same selector works here. But we can't assume that the package contents view and the covered eresources view will stay the same. In fact, they'll probably be different very soon, when we add the ability to perform searches on the list in the Package Contents section.

test/bigtest/tests/package-content-test.js
aditya-matukumalli: No specific reason. Will just change the interactor variables to camelCase

test/bigtest/tests/package-content-test.js
aditya-matukumalli: Yes. Added a `CoveredEresourcesListInteractor` within `AgreementViewInteractor` that has the clicks on the filters.

PR #344, opened 2019-11-04
PR for ERM-544

EthanFreestone:
The warning that the status of an amendment on a linked license inside an agreement conflicts with the status of that license amendment is now using the MessageBanner component.


src/components/LicensesFieldArray/AmendmentsFieldArray.js
```diff
@@ -78,18 +73,18 @@ class AmendmentsFieldArray extends React.Component {
       license = {},
       name
     } = this.props;
+    const { statusWarnings } = this.state;
     const { amendments = [] } = license;
-
-    if (!items.length) return null;
-
+    if (!items.length) {
```
doytch: Do you have an automatic code formatter enabled or something? Might want to leave it off since it ended up making this section a bit less readable.

src/components/LicensesFieldArray/AmendmentsFieldArray.js
```diff
@@ -35,40 +35,35 @@ class AmendmentsFieldArray extends React.Component {
...
+        for (let i = 0; i < props.items.length; i++) {
+          this.state.statusWarnings[i] = { warningId: '', warningValuesStatus: null };
+        }
       }
+    }
+  }
 
-      form.mutators.setFieldData(`${name}[${i}].status`, { warning });
-    });
+  RenderWarningStatus(id, valuesStatus) {
```
doytch: Class methods should be camelCased, eg, `renderWarnings()`

src/components/LicensesFieldArray/AmendmentsFieldArray.js
```diff
@@ -131,14 +126,42 @@ class AmendmentsFieldArray extends React.Component {
...
+                              warning = 'ui-agreements.license.warn.amendmentPast';
+                            }
+                            // Amendment has an invalid status.
+                            const linkedStatus = get(amendment, 'status', {});
+                            if (linkedStatus.value === statuses.EXPIRED || linkedStatus.value === statuses.REJECTED) {
+                              warning = 'ui-agreements.license.warn.amendmentStatus';
+                              warningValuesStatus = linkedStatus.label;
+                            }
+                          }
+                          statusWarnings[i].warningId = warning;
```
doytch: This isn't how you set state for a component. See [the docs](https://reactjs.org/docs/react-component.html#setstate) or [this CSS tricks article](https://css-tricks.com/understanding-react-setstate/) (since the docs are down while I'm writing this).

src/components/LicensesFieldArray/AmendmentsFieldArray.js
doytch: Even knowing what this PR is about, I'm not really certain what valuesStatus is going to be. After digging around I see it's the amendment record's status, which doesn't really fit this variable name. The same goes for the various other permutations of valuesStatus throughout this PR.

doytch: Finally, I'm not really sold on the method of this function. I think the whole PR would be clearer if we cut it out.

src/components/views/AgreementForm.js
```diff
@@ -178,7 +178,6 @@ class AgreementForm extends React.Component {
     const { form, initialValues: { id, name } } = this.props;
 
     const hasLoaded = form.getRegisteredFields().length > 0;
-
```
doytch: Let the code breathe, don't unnecessarily delete empty lines. :)

src/components/LicensesFieldArray/AmendmentsFieldArray.js
```diff
@@ -35,40 +35,35 @@ class AmendmentsFieldArray extends React.Component {
     name: PropTypes.string.isRequired,
   };
 
-  warnStatusMismatch = ({ values }) => {
-    const { form, name } = this.props;
+  constructor(props) {
+    super(props);
+    this.state = { statusWarnings: [] };
```
doytch: So, per the comments in the review body, this isn't necessary at all, but if it _was_, you could write this a lot cleaner via something like the following:
```
const { items = [] } = props; // guarantees items exists and is an empty array so we can perform array methods on it

this.state = {
  warnings = items.map(() => ({ warningId: '', warningValuesStatus: null }))
}
```
There aren't a tonne of situations where a for-loop is actually the preferred way of writing a loop in JS. 

Also, falsey values can be a really tricky situation in JS, so avoid `null` except where you're _definitely_ sure you specifically want a `null` as opposed to the much more common `undefined`. Empty strings are useful for when you want a falsey value to call string methods on a variable without checking if it's a string (vs undefined) or if you're using it in a template literal. Ie, ``` `${foo}` ``` => `"undefined"` if foo is undefined, rather than just an empty string.

src/components/LicensesFieldArray/AmendmentsFieldArray.js
EthanFreestone: Yep, this makes sense. I wasn't aware that we were certain that the order would never change, hence why I went to the extra effort. This is a lot cleaner though. I wasn't aware that we could set, say, Array[2] = "whatever" on a potentially empty array... presumably there's some magic happening in the background there to make the Array [undefined, undefined, "whatever"]...? Also, for some reason the implicit setting of `warnings` in `this.setState` didn't work, so I changed it to be `return {warnings: warnings}`.

src/components/LicensesFieldArray/AmendmentsFieldArray.js
EthanFreestone: Yep, agreed. I wasn't super sold on this either, it seemed hacky to me, and I didn't really know what to call it. At the time it was the only way I seemed to be able to get it to work. Now I know that orders aren't changing and we can just use a simple array, this whole method has been removed.

src/components/views/AgreementForm.js
EthanFreestone: I don't remember doing this, I've added it back in now.

src/components/LicensesFieldArray/AmendmentsFieldArray.js
EthanFreestone: Duly noted, as above this has now been removed anyway.

src/components/LicensesFieldArray/AmendmentsFieldArray.js
EthanFreestone: Yeah, I'm kinda surprised this worked looking back... code has been changed to remedy this.

src/components/LicensesFieldArray/AmendmentsFieldArray.js
```diff
@@ -131,14 +100,47 @@ class AmendmentsFieldArray extends React.Component {
...
                   <Field
-                    component={Select}
-                    dataOptions={amendmentStatusValues}
-                    label={<FormattedMessage id="ui-agreements.license.prop.status" />}
                     name={`${name}[${i}].status`}
-                    placeholder=" "
-                    required
                     validate={validators.required}
-                  />
+                    label={<FormattedMessage id="ui-agreements.license.prop.status" />}
```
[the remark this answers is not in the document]
EthanFreestone: Ah ok, I shall do this

PR #356, opened 2019-11-20
ERM-558: erm/sas/<uuid>/export options not working in all cases

CalamityC:
* add tooltip
* update language files en.json
* disable export button if filterPath is 'dropped' or 'future'


translations/ui-agreements/en_US.json
```diff
@@ -203,6 +203,7 @@
     "eresourcesCovered.exportAsKBART": "Export as KBART",
     "eresourcesCovered.json": "JSON",
     "eresourcesCovered.kbart": "KBART",
+    "eresourcesCovered.exportButton.tooltip": "Exports only supported for current or all resources",
```
doytch: Don't make _additions_ to translation files other than `en`. It's not necessary because of the way the fallbacks work and it breaks the Localize workflow.

It is however fine to make _changes_ to existing keys in en_US so that the correct string shows up.

src/components/AgreementSections/CoveredEResourcesList.js
```diff
@@ -106,49 +106,78 @@ export default class CoveredEResourcesList extends React.Component {
...
+      >
+        {disabled ?
+          <Tooltip
+            id="covered-eresources-export-tooltip"
+            placement="top"
+            text={<FormattedMessage id="ui-agreements.eresourcesCovered.exportButton.tooltip" />}
+          >
+            {({ ref, ariaIds }) => (
+              <div
+                ref={ref}
```
doytch: You should be setting the ref and aria attribute [on the DropdownButton itself](https://github.com/folio-org/stripes-components/blob/78d983eea6473b92116f8c7bfac3a85c1853068e/lib/DropdownButton/DropdownButton.js#L14), and not on a wrapper `div`.

src/components/AgreementSections/CoveredEResourcesList.js
```diff
@@ -106,49 +106,78 @@ export default class CoveredEResourcesList extends React.Component {
...
+                aria-labelledby={ariaIds.text}
```
doytch: Additionally, you should [set the `aria-describedby` attribute](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute) and not `aria-labelledby`. You're describing the button with this tooltip, not labelling it; the thing that labels the button is the _actual_ label of the button.

src/components/AgreementSections/CoveredEResourcesList.js
CalamityC: I tried that first but I couldn't  get it working. The tooltip doesn't show up if I write the following
```
<Tooltip
            id="covered-eresources-export-tooltip"
            placement="top"
            text={<FormattedMessage id="ui-agreements.eresourcesCovered.exportButton.tooltip" />}
>
            {({ ref, ariaIds }) => (
               <DropdownButton
                  data-role="toggle"
                  disabled={disabled}
                  ref={ref}
                  aria-describedby={ariaIds.text}
               >
                   <FormattedMessage id="ui-agreements.eresourcesCovered.exportAs" />
              </DropdownButton>
           )}
</Tooltip>
```

doytch: So the key things are setting the aria- and ref props correctly. If you've confirmed the aria-describedby is being set, then that leaves checking ref.

If you set `ref={ref}` on DropdownButton, then your ref is for the React component itself. You need it to be on the rendered `<button>` (lowercase). So if you look at DropdownButton's code, you see it passes down most props to Button. If you check Button's code, you'll see it allows you to set the ref of the rendered `<button>` if you give it a `buttonRef` prop.

So that'd be the next thing I'd try.

CalamityC: adding a `buttonRef` prop to the DropdownButton doesn't work either

doytch: Hmm, looks like this is actually an issue with [the behaviour of `disabled` Buttons not firing events](https://github.com/folio-org/stripes-components/blob/master/lib/Button/Button.css#L218). This is an issue for us since `Tooltip` relies on the `<button>` [sending `mouseover` events to decide whether to show the tooltip or not](https://github.com/folio-org/stripes-components/blob/master/lib/Tooltip/Tooltip.js#L57).

@rasmuswoelk, do you have a suggestion on what the best course of action is here? Should we go ahead and just wrap the `DropdownButton` in a div like it is, or something else? Turn on `pointer-events` in our CSS for this instance?

rasmuswoelk: @doytch Hmm. Good question.

In general, we need Tooltips to be accessible both on hover and focus to comply with the accessibility standards. So when we make the button disabled, it's no longer possible to focus it so in that case, it's actually the intended behavior here (the tooltip not showing on disabled buttons).

Could we solve this by placing an`<Icon tabIndex={0} />` next to the dropdown, which will render the tooltip and have that provide this information for the user?

image

doytch: Hmmm, that looks pretty nice...

@CalamityC lets go with this approach. It even has the benefit of cleaning up the render code!

PR #357, opened 2019-11-20
ERM-463 and ERM-584: Linked License Amendments additions

EthanFreestone:
This is a PR for the issues ERM-463 and ERM-584. This makes the following changes:

- Notes have been moved from an infopopover to a column in the end of the table (awaiting a stripes component that would allow a "show more" kind of approach)
- A new 'warnings' column has taken its place, with a warning icon appearing if a status conflict is observed, with a tooltip containing the warning itself
- A banner displays if there are unassigned amendments on the controlling license, and a table displays these at the bottom of the accordion
- Props added to LicenseAmendmentList to track whether to display notes and warnings, and the logic for visibility of columns has been changed slightly to accommodate this.


src/components/AgreementSections/ControllingLicense.js
```diff
@@ -83,6 +91,18 @@ export default class ControllingLicense extends React.Component {
               id="controlling-license-current-amendments"
               license={licenseRecord}
               renderStatuses
+              renderWarnings
+              renderNotes
+            />
+          </KeyValue>
+          : null
+        }
+        { unsetAmendments.length ?
```
doytch: Why can't we just `this.renderAmendments` here? It seems like all we're doing is moving it into the card. Can tweak that method if it doesn't do exactly what you need.

src/components/AgreementSections/ControllingLicense.js
```diff
@@ -105,6 +125,7 @@ export default class ControllingLicense extends React.Component {
           id={`controlling-license-${statusString}-amendments`}
           license={linkedLicense.remoteId_object}
           renderStatuses={status !== statuses.HISTORICAL}
+          renderNotes={status !== null}
```
doytch: I _believe_ you're trying to render notes unless the notes are unset. If so, maybe make it more obvious by rewriting this as 

```suggestion
          renderNotes={statusString !== status.UNSET}
```

And while you're at it, change that `renderStatuses` line to check `statusString` instead of `status` as well, for uniformity.

doytch: It may also just be nicer to go stop relying on this dumb `null -> 'unset'` translation that I do here...

Like, I wonder if you could just go into getLicenseAmendments.js:23 and change the default value from `null` to `statuses.UNSET` (after importing `statuses`, naturally).

Then, change the places where I pass `null` into that function or rely on `null` and instead use `statuses.UNSET`... It'd also let you ditch that `statusString` vs `status` mess in this function...

That'd be ideal I think. If doing that clean up makes you uncertain or you start trying and it messes everything up, it's fine to skip it. I can pick it up afterward.

src/components/LicenseAmendmentList/LicenseAmendmentList.css
```diff
@@ -1,3 +1,9 @@
+@import "@folio/stripes-components/lib/variables";
+
 .note {
```
doytch: Looks like you're not using this set anymore so should delete it.

src/components/LicenseAmendmentList/LicenseAmendmentList.js
```diff
@@ -26,6 +27,46 @@ export default class LicenseAmendmentList extends React.Component {
       id: PropTypes.string.isRequired,
     }),
     renderStatuses: PropTypes.bool,
+    renderWarnings: PropTypes.bool,
+    renderNotes: PropTypes.bool,
+  }
+
+  renderStatusMismatchWarnings(amendment) {
```
doytch: This is pretty cluttered and can be cleaned up. Some examples:
- Initing everything to `null` isn't necessary. Not only are you not comparing to `null`, `null` sets of warnings in the heads of JS devs like me since you only use it if you absolutely want _null_ and not `undefined` or `false` or `0` or `""`.
  - Use `get()` for multi-level indexing like `const status = get(amendment, 'status.value')`.
  - Use destructuring for simple ones like `const { startDate, endDate } = amendment`
- The explicit `return null`s really clutter things up here. It makes it really hard to track what's actually going on and what's necessary and what isn't. Consider something like the following:

```
if (agreementStatus === statuses.CURRENT) {
  if (licenseStatus === statuses.EXPIRED || licenseStatus === statuses.REJECTED) {
    return <FormattedMessage />;
  } else if (startDate && new Date(amendment.startDate).getTime() > new Date().getTime()) {
    return <FormattedMessage />;
  } else if (endDate && new Date(amendment.endDate).getTime() < new Date().getTime()) {
    return <FormattedMessage />;
  }
}

return null;
```
- Finally, the variable names are really confusing. `licenseStatus` is not the status of the license and `agreementStatus` is not the status of the agreement.

doytch: Also...I thought this logic seemed familiar and it looks like [we already do this kind of thing here](https://github.com/folio-org/ui-erm/blob/5cda051fa068dcf7881338507f976d74f5ac4f98/src/components/LicensesFieldArray/AmendmentsFieldArray.js#L115). Is there a way we could just generalise the validation, put it in `utilities/validators.js`, and call that validator both in that spot and in here?

src/components/AgreementSections/ControllingLicense.js
EthanFreestone: We can do that here, I was simply following the example above, where currentAmendments are rendered directly. Of course I can also make that part use `this.renderAmendments`, which if I'm already tweaking that method I think I will do, as little repeated code as possible.

src/components/AgreementSections/ControllingLicense.js
EthanFreestone: This makes sense, I've changed the default value for an unassigned amendment to be statuses.UNSET, and the code now relies on that, removing the need for statusString at all (I believe)

src/components/LicenseAmendmentList/LicenseAmendmentList.css
EthanFreestone: Which set is this sorry? The note set I didn't change anything, and the import of variables is used below for var(--warn), so I'm a little confused.

EthanFreestone: I couldn't figure out how to use the `.note` styling to affect <div>, so I stuck to the explicit column width method. I've deleted that `.note` bit now, let me know if that's not what you meant

src/components/LicenseAmendmentList/LicenseAmendmentList.js
EthanFreestone: Thank you for reminding me of the neater way to do that kind of thing, I hated the way it looked with all those `else { return.. }` bits, but couldn't get my head around it, I'll definitely refactor that. As for the generalisation, I definitely can do that, but would validators be the right place for it? It's kind of it's own thing, warnings. In addition, talking to Gill it sounded like there is a possibility that the list of places we need warnings for might expand in future, so it might be good to have a separate file in `utilities` just for warnings at some point? For now I've just put them in validators, but it might be worth keeping in mind.

doytch: I guess `validators` is a bit of a prescriptive term to use in the file name. Really, the only thing that makes them a true final-form validator is that you assign them to the `validator` prop. If you use them in the context of a warning (as I believe we do in AmendmentsFieldArray), then yeah, they're not really functioning as a truuuue final-form validator. I just called it `validators.js` because they follow the RFF pattern, but you definitely don't **have** to use it as a way of permitting/denying form submission.

That's all kind of hand-wavy...I've rewritten that above paragraph a few times so lemme know if it's not clear.

doytch: Yeah I see your point, I think you're right. Especially since it doesn't have the same function signature as a traditional validator. I'm not against a `utilities/getAmendmentStatusConflicts.js` or something instead.

src/components/LicensesFieldArray/AmendmentsFieldArray.js
```diff
@@ -111,23 +110,8 @@ class AmendmentsFieldArray extends React.Component {
...
-                            }
-
-                            // Amendment has an invalid status.
-                            const linkedStatus = get(amendment, 'status', {});
-                            if (linkedStatus.value === statuses.EXPIRED || linkedStatus.value === statuses.REJECTED) {
-                              warning = <FormattedMessage id="ui-agreements.license.warn.amendmentStatus" values={{ status: linkedStatus.label }} />;
-                            }
-                          }
+                          const linkedStatus = get(amendment, 'status', {});
+                          const warning = validators.amendmentWarning(value, linkedStatus.value, linkedStatus.value, amendment.startDate, amendment.endDate);
```
doytch: Third param should be `linkedStatus.label` no?

src/components/utilities/validators.js
```diff
@@ -109,10 +109,24 @@ const overlappingDates = (value, allValues, meta, errorMessageKey) => {
   return undefined;
 };
 
+const amendmentWarning = (statusInAgreement, statusInLicense, statusInLicenseLabel, startDate, endDate) => {
```
doytch: This is one of those functions that I'm afraid is gonna get tweaked and updated and messed with over and over again. People will ask that we check this field and that field, etc. And then we're gonna get stuck tacking on more and more things as function parameters.

So in cases like this, I'd prefer to just take a single object that has keys for the various things were checking, since they're all related to the amendment after all. So something like:

```
const amendmentWarning = amendment => {
  const {
    statusInAgreement,
    statusInLicense,
    startDate,
    endDate,
  } = amendment;

  if (statusInAgreement.value === statuses.CURRENT) {
    if (statusInLicense.value === statuses.EXPIRED) {
    ....
```

I went ahead and left the two `status...` entries as an object with `{ value, label }` entries because it's:
1. More flexible down the road. (Let's say we now want to use the statusInAgreement.label...)
1. Is cleaner when _calling_ this function. I know it dirties up this function a bit more, but in cases where it's six-of-one-half-dozen-of-another like this, I err on the side of cleaning up the calling experience since that'll be done more often.

src/components/AgreementSections/ControllingLicense.js
```diff
@@ -52,7 +52,7 @@ export default class ControllingLicense extends React.Component {
 
   renderLicense = linkedLicense => {
     const currentAmendments = getLicenseAmendments(linkedLicense, statuses.CURRENT);
-    const unsetAmendments = getLicenseAmendments(linkedLicense, null);
+    const unsetAmendments = getLicenseAmendments(linkedLicense, statuses.UNSET);
```
doytch: What this line is doing is so much more obvious now! 😍

src/components/utilities/getLicenseAmendments.js
```diff
@@ -20,7 +21,7 @@ export default (linkedLicense, status) => {
     });
 
   if (status !== undefined) {
-    return amendments.filter(a => get(a, 'statusForThisAgreement.value', null) === status);
+    return amendments.filter(a => get(a, 'statusForThisAgreement.value', statuses.UNSET) === status);
```
doytch: Yeah this is way more obvious, I love it. Thanks!

src/components/AgreementSections/ControllingLicense.js
doytch: Hmmm, yeah good catch. I'm honestly not sure why I didn't use `renderAmendments` for the current amendments and instead wrote it out. Nothing stands out. You're right there's a tiny performance hit but the benefit in terms of making the code easier to maintain is big enough that it's worth it IMO.

src/components/LicensesFieldArray/AmendmentsFieldArray.js
EthanFreestone: Whoops, it should indeed

src/components/utilities/validators.js
EthanFreestone: This is fair, that's a way cleaner approach. I think I agree that it's better to clean up the calling code. 

The issue is what happens when we use it in a situation where we want to use data that isn't necessarily in an amendment. In `AmendmentsFieldArray` for instance, we're calling it using `e.target` for the `statusInAgreement`, and that value obviously wouldn't be passed in with just `amendment`.

For now I can make it take an amendment and the `statusInAgreement` separately, but this feels a bit like a dirty compromise, I'd be interested to know if there was a cleaner way still.

src/components/AgreementSections/ControllingLicense.js
EthanFreestone: Cheers :)

src/components/utilities/getLicenseAmendments.js
EthanFreestone: Nice one 😄

src/components/AgreementSections/ControllingLicense.js
```diff
@@ -77,14 +85,11 @@ export default class ControllingLicense extends React.Component {
...
         { currentAmendments.length ?
-          <KeyValue label={<FormattedMessage id="ui-agreements.license.currentAmendments" />}>
-            <LicenseAmendmentList
-              amendments={currentAmendments}
-              id="controlling-license-current-amendments"
-              license={licenseRecord}
-              renderStatuses
-            />
-          </KeyValue>
+          this.renderAmendments(linkedLicense, statuses.CURRENT)
```
doytch: Looking at this...doesn't it seem as though we can just skip the generation of `currentAmendments` in this function and just always call `this.renderAmendments(linkedLicense, statuses.CURRENT)`? I don't really see where we're using `currentAmendments`

src/components/LicenseAmendmentList/LicenseAmendmentList.js
```diff
@@ -26,40 +27,74 @@ export default class LicenseAmendmentList extends React.Component {
...
+                text={this.renderStatusMismatchWarnings(a)}
+                placement="left"
+              >
+                {({ ref, ariaIds }) => (
+                  // For the time being I'm using the workaround of a <span> while we wait to see what can/should be added to the 'Icon' component
+                  <Icon
+                    icon="exclamation-circle"
+                    iconClassName={css.tooltipIcon}
+                    ref={ref}
+                    aria-label={ariaIds.text}
```
doytch:
```suggestion
                    aria-labelledby={ariaIds.text}
```
`ariaIds` provides `id`s of the elements that will provide the label/description, not the label/description itself. So we need to pass the id to [`aria-labelledby`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-labelledby_attribute) (or `aria-describedby` in other cases), not `aria-label`.

src/components/AgreementSections/ControllingLicense.js
EthanFreestone: We're using it right now to check if there are any or not, to check whether to render ```{ currentAmendments.length ?
          this.renderAmendments(linkedLicense, statuses.CURRENT)
          : null
        }```

src/components/LicenseAmendmentList/LicenseAmendmentList.js
EthanFreestone: Ah ok, change made

src/components/AgreementSections/ControllingLicense.js
doytch: Right but check the definition of `renderAmendments` - if we call it unconditionally it'll do the same thing, right?

src/components/utilities/validators.js
doytch: Yeah I'd probably mix it into the amendment itself. Eg,
```
amendmentWarning({
  ...amendment,
  statusInAgreement: e.target.value,
})
```

src/components/AgreementSections/ControllingLicense.js
EthanFreestone: Oh, damn, you're right. Same goes for `unsetAmendments` then I guess

src/components/utilities/validators.js
EthanFreestone: Right, but if `getConflictWarnings` has code at the beginning to set `statusInAgreement` for all of those other times, then won't it always get overwritten? Unless I should be putting a check on each one to say "if this exists already, then just use that"?

src/components/AgreementSections/ControllingLicense.js
doytch: Yeah I think it's fine to leave it as-is for the unset amendments. It's more obvious what we're doing there then.

src/components/utilities/validators.js
doytch: I wasn't clear, that snippet in my last message is how you'd _call_ `amendmentWarning`. So you create a new object (the amendment), spreading in all the existing amendment properties and then adding the `statusInAgreement` property.

EthanFreestone: Ohhh I just realised you meant I could overwrite the variable, which inside the amendment is `statusForThisAgreement`, then just always have `getConflictWarnings` grab that variable

EthanFreestone: Oh, I guess I was still off then. So I DO need `getConflictWarnings` to check whether it's being passed a `statusInAgreement`, and then if it isn't then grab `statusForThisAgreement` instead

src/components/LicenseAmendmentList/LicenseAmendmentList.js
```diff
@@ -26,40 +26,73 @@ export default class LicenseAmendmentList extends React.Component {
...
-          note: a => (a.note ? <InfoPopover contentClass={css.note} content={a.note} /> : ''),
+          warning: a => (
+            this.renderStatusMismatchWarnings(a) ?
+              <Tooltip
+                id={`warning-tooltip-${a.id}`}
+                text={this.renderStatusMismatchWarnings(a)}
+                placement="left"
+              >
+                {({ ref, ariaIds }) => (
+                  // For the time being I'm using the workaround of a <span> while we wait to see what can/should be added to the 'Icon' component
```
doytch:
```suggestion
```

doytch:
We're still probably gonna need integration tests for this (likely an extension to the `linked-licenses` test), but unfortunately we're hitting issues with our integration tests because of https://github.com/final-form/final-form/issues/300

I'm gonna sit on this PR to see how that issue proceeds.

doytch:
OK yeah, we're pegging FF and RFF to earlier versions in our builds atm so tests should and can actually be written for this.

To get the older versions in your env, update your workspace's `package.json` to have a `resolutions` entry like the following:
```
{
  "private": true,
  "workspaces": ["*"],
  "dependencies": {},
  "resolutions": {
    "final-form": "4.18.5",
    "react-final-form": "6.3.0"
  }
}
```

test/ui-testing/linked-license-warnings.js
```diff
@@ -0,0 +1,370 @@
...
+            amendments.forEach(a => {
+              chain = chain
+                .wait('#clickable-expand-all')
+                .click('#clickable-expand-all')
+                .click('#add-amendment-button')
+                .wait('#edit-amendment-name')
+                .insert('#edit-amendment-name', a.name)
+                if (a.status) {
+                  chain = chain
+                    .wait('#edit-amendment-status')
```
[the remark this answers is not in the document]
EthanFreestone: Ah right, that's fair. I shall bear it in mind. But if it does defend against refactors in future then I'll probably just leave it be for the time being :)

test/ui-testing/linked-license-warnings.js
```diff
@@ -0,0 +1,370 @@
...
+    AMENDMENT_NAME_INDEX: 1,
+    NOTE_INDEX: 5
+  };
+  
+  const checkWarningColumn = (nightmare, table, shouldExist, done) => {
+    let chain = nightmare
+    .wait('#clickable-expand-all')
+    .click('#clickable-expand-all')
+    .wait(`${table}`)
+    .evaluate((table, shouldExist) => {
```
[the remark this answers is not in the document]
EthanFreestone: Done!

PR #358, opened 2019-11-20
Error handling for cloning an agreement

aditya-matukumalli:


src/routes/AgreementViewRoute.js
```diff
@@ -226,21 +226,31 @@ class AgreementViewRoute extends React.Component {
...
-        history.push(`${urls.agreementEdit(id)}${location.search}`);
-      });
+    });
+
+    try {
+      const text = await response.text(); // Parse it as text
+      const data = JSON.parse(text); // Try to parse it as json
+      if (response.ok) {
+        if (data.id) history.push(`${urls.agreementEdit(data.id)}${location.search}`);
+        else throw new Error('invalidJsonError'); // when the json response body doesn't contain an id
```
doytch: Don't use magic strings, add a new file to `constants/`. `errorTypes` perhaps? Dunno.

aditya-matukumalli: Okay. Added them in to the constants

src/constants/errorTypes.js
```diff
@@ -0,0 +1,4 @@
+export default {
+  INVALID_JSON_ERROR: 'invalidJsonError',
+  JSON_ERROR: 'jsonError'
```
doytch: It looks like you don't use this type?

src/routes/AgreementViewRoute.js
```diff
@@ -226,21 +227,31 @@ class AgreementViewRoute extends React.Component {
...
-    }).then(response => response.json())
-      .then(({ id }) => {
-        history.push(`${urls.agreementEdit(id)}${location.search}`);
-      });
+    });
+
+    try {
+      const text = await response.text(); // Parse it as text
+      const data = JSON.parse(text); // Try to parse it as json
+      if (response.ok) {
```
doytch: If you're not using JSON_ERROR, you can simplify this section to just `if (response.ok && data.id) ....`

PR #360, opened 2019-11-22
PR for ERM-585: Added hasMargin to fix some small graphical inconsistencies

EthanFreestone:
This PR fixes about half of ERM-585, the rest will be fixed by a PR in erm-stripes-components


EthanFreestone:
This should not be merged, @doytch has recommended that we push all of these changes instead into `EditCard` itself, so that `hasMargin` is always passed

PR #376, opened 2019-12-09
ERM-581: Small consistency changes to translations for buttons in edit pane

EthanFreestone:


doytch:
Can you update en_US with these changes? It's confusing, but the one exception to the rule of "only update en.json" is when you're editing an existing translation string. In that case, make the change in en.json and en_US.json. 

This is because otherwise it won't be QAable until the next translation bot run otherwise.

PR #382, opened 2019-12-11
Refactor FieldArrays and apply focus on the first field on initial m…

aditya-matukumalli:
…ount, refs ERM-575

The various `FieldArray` components are refactored to factor out the `Field`s into their own component , so that we could just handle the focus easily now on mounting of a new `Field` instead of complicating this inside a  `FieldArray` component

### DIFF
--- a/translations/ui-agreements/en.json
+++ b/translations/ui-agreements/en.json
@@ -98,6 +98,12 @@
   "agreementLines.hasCustomCoverage": "Has custom coverage",
   "agreementLines.customCoverageTitle": "Custom coverage #{number}",
 
+  "agreementLines.customCoverage.startIssue": "Start issue",
+  "agreementLines.customCoverage.endIssue": "End issue",
+  "agreementLines.customCoverage.startVolume": "Start Volume",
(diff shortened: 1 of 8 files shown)


translations/ui-agreements/en.json
```diff
@@ -98,6 +98,11 @@
   "agreementLines.hasCustomCoverage": "Has custom coverage",
   "agreementLines.customCoverageTitle": "Custom coverage #{number}",
 
+  "agreementLines.customCoverage.startIssue": "Start issue",
+  "agreementLines.customCoverage.endIssue": "End issue",
+  "agreementLines.customCoverage.startVolume": "Start Volume",
```
doytch:
```suggestion
  "agreementLines.customCoverage.startVolume": "Start volume",
```

translations/ui-agreements/en.json
```diff
@@ -98,6 +98,11 @@
...
+  "agreementLines.customCoverage.endVolume": "End Volume",
```
doytch:
```suggestion
  "agreementLines.customCoverage.endVolume": "End volume",
```

PR #385, opened 2019-12-16
PR For ERM-604

EthanFreestone:
Changes made to usage data provider card to bring it into line with organisations card


src/components/UsageDataProvidersFieldArray/UsageDataProviderField.js
```diff
@@ -63,16 +63,20 @@ export default class UsageDataProviderField extends React.Component {
...
   renderUDP = () => {
     const { udp } = this.props;
-
     return (
-      <div data-test-udp-card-name>
-        <AppIcon app="erm-usage" size="small">
-          <strong>{udp.label}</strong>
-        </AppIcon>
-      </div>
+      <Row>
```
aditya-matukumalli: Any specific reason why you are rendering the `Row` and `Col` here? Like could you just not render `KeyValue` as is?

PR #396, opened 2020-01-07
PR for ERM-629

EthanFreestone:
This deals with the ui-agreements half of ERM-629, including changes to allow indexing of the "remove" button tooltips, and adding a tooltip with specific information for "replace" buttons


src/components/UsageDataProvidersFieldArray/UsageDataProviderField.js
```diff
@@ -41,18 +41,40 @@ export default class UsageDataProviderField extends React.Component {
       onUDPSelected={this.props.onUDPSelected}
       renderTrigger={(props) => {
         this.triggerButton = props.buttonRef;
+        const UPDName = this.props.udp ? this.props.udp.label : '';
```
doytch:
```suggestion
        const UPDName = get(this.props.udp, 'label');
```

EthanFreestone: This, and the conditional tooltip rendering changes are now done

PR #397, opened 2020-01-08
Add tooltips for buttons that add/remove title/packages from the bask…

aditya-matukumalli:
…et, refs ERM-640

Added tooltips for the add/remove package/title buttons under the Options for acquiring e-resource accordion


src/components/AddToBasketButton/AddToBasketButton.js
```diff
@@ -74,24 +78,62 @@ class AddToBasketButton extends React.Component {
...
+          >
+            {({ ref, ariaIds }) => (
+              <IfPermission perm="ui-agreements.agreements.edit">
+                <Button
+                  {...buttonProps}
+                  aria-labelledby={ariaIds.text}
+                  buttonStyle={itemExistsInBasket ? 'default' : 'primary'}
+                  data-test-basket-add-button={itemExistsInBasket ? undefined : true}
+                  data-test-basket-remove-button={itemExistsInBasket ? true : undefined}
+                  data-test-coverage-details={coverage}
```
doytch: Can we clean up all these props that are duplicated across the two render modes (w/ Tooltip and w/o)? Ie, create a `baseProps` variable and spread that object into the `Button`'s props.

(I just merged this [eslint-config-stripes PR](https://github.com/folio-org/eslint-config-stripes/pull/54/files) so if you re-yarn in the new build, you won't need to quote all the props, just the ones with hyphens)

doytch: Also, I think you can wrap the entire thing in an `IfPermission` instead, so you don't have to define that twice here as well.
```
const button = addButtonTooltipText ? <Tooltip /> : <Button />;
return <IfPermission>{button}</IfPermission>;
```

aditya-matukumalli:
Requested changes fixed here -> https://github.com/folio-org/ui-agreements/compare/erm-640-cleanup?expand=1

PR #404, opened 2020-01-15
Migrate eresources list to use new stripes-connect paging functionali…

aditya-matukumalli:
…ty, refs ERM-642

Previously when we first render the eresources list, it rendered all the pages at once, generally in th e order of approx: N*10K records at once, but with the new paging functionality, we can customize the number of records that are fetched on the initial render and the number of records to be fetched on subsequent scrolls. This PR fetches 100 records at a time per scroll. Similar PR [here](https://github.com/folio-org/ui-local-kb-admin/pull/70) to handle error/info logs on a job in the local-kb-admin app


src/components/views/EResources.js
```diff
@@ -313,6 +313,8 @@ export default class EResources extends React.Component {
                       onNeedMoreData={onNeedMoreData}
                       isSelected={({ item }) => item.id === selectedRecordId}
                       onRowClick={onSelectRow}
+                      pageAmount={100}
```
doytch: If this maps to the RESULT_COUNT_INCREMENT in EResourcesRoute, can we move that constant to `constants/` and import it from there in here and EResourcesRoute?

src/routes/EResourcesRoute.js
```diff
@@ -9,17 +9,17 @@ import { getSASParams } from '@folio/stripes-erm-components';
...
 
 class EResourcesRoute extends React.Component {
   static manifest = Object.freeze({
     eresources: {
       type: 'okapi',
       path: 'erm/resource/electronic',
       records: 'results',
-      recordsRequired: '%{resultCount}',
+      resultOffset: '%{resultOffset}',
       perRequest: 100,
```
doytch: Same as the other PR
```suggestion
      perRequest: resultCount.RESULT_COUNT_INCREMENT,
```

aditya-matukumalli:
@doytch Added in the paging functionality within e-resources in package

src/routes/EResourceViewRoute.js
```diff
@@ -39,18 +40,19 @@ class EResourceViewRoute extends React.Component {
...
         sort: 'pti.titleInstance.name;asc',
         stats: 'true',
       },
     },
     query: {},
-    entitlementsCount: { initialValue: RECORDS_PER_REQUEST },
+    entitlementsCount: { initialValue: resultCount.INITIAL_RESULT_COUNT },
     packageContentsFilter: { initialValue: 'current' },
-    packageContentsCount: { initialValue: RECORDS_PER_REQUEST },
+    resultOffset: { initialValue: 0 },
```
doytch: Can we be a bit more specific about what `result` we mean in `resultCount` and `resultOffset`?  We've got a pair of these resources that need these kinds of variables right now (entitlements is the other one) and could have more in the future so it can be confusing.
```suggestion
    packageContentsOffset: { initialValue: 0 },
```

src/routes/EResourceViewRoute.js
```diff
@@ -135,6 +132,17 @@ class EResourceViewRoute extends React.Component {
...
+    mutator.resultOffset.replace(0);
+    mutator.resultCount.replace(resultCount.INITIAL_RESULT_COUNT);
+  }
+
+  handleNeedMorePackageContents = (_askAmount, index) => {
+    const { mutator, resources } = this.props;
+    if (index > 0) {
+      mutator.resultOffset.replace(index);
+    } else {
+      mutator.resultCount.replace(resources.resultCount + resultCount.RESULT_COUNT_INCREMENT);
```
doytch: Does this case ever run? Aren't we always using the offset-fetch, so this onNeedMore callback will always have an `index`?

src/routes/EResourceViewRoute.js
```diff
@@ -181,15 +192,18 @@ class EResourceViewRoute extends React.Component {
...
         handlers={{
           ...handlers,
           onFilterPackageContents: this.handleFilterPackageContents,
+          onNeedMorePackageContents: this.handleNeedMorePackageContents,
           onClose: this.handleClose,
           onToggleTags: tagsEnabled ? this.handleToggleTags : undefined,
         }}
         helperApp={this.getHelperApp()}
         isLoading={this.isLoading()}
+        isPending={get(resources, 'packageContents.isPending', true)}
```
doytch: I'm really not a fan of this.

1. It's a really generic prop name that you pass several levels deep. "_What's_ pending?"
1. It introduces another way of checking for whether a resource exists or not. Ultimately the purpose of this prop is so that we can [show a Spinner if a new filter is selected](https://github.com/folio-org/ui-agreements/pull/404/files#diff-118d7b474136e214559b4dd3b6fb33d9R139), right? Normally, when we need to a show a Spinner on initial load, we just check if the records are `undefined` (as opposed to an empty array). [Note the intentional lack of a default `[]` param for the `get()`.](https://github.com/folio-org/ui-erm/blob/bb06797ee945ecf13f2a866c4770ec8dddb33ebb/src/routes/AgreementViewRoute.js#L214)

So I'd say let's avoid passing any kind of `isPending` prop here. Instead, when `packageContents` is fetching the first page (upon initial load or after the filter is changed), pass its records as `undefined`. That may require tweaks to `this.getRecords`, and those tweaks may require https://github.com/folio-org/stripes-connect/pull/124.

src/components/EResourceSections/PackageContents.js
```diff
@@ -24,11 +25,23 @@ export default class PackageContents extends React.Component {
...
+    isPending: PropTypes.bool,
     onFilterPackageContents: PropTypes.func.isRequired,
+    onNeedMorePackageContents: PropTypes.func.isRequired,
     onToggle: PropTypes.func,
     open: PropTypes.bool,
   };
 
+  state = {};
+
+  static getDerivedStateFromProps(props, state) {
```
doytch: Yeah I think I just became more in favour of https://github.com/folio-org/stripes-connect/pull/124 if it lets us avoid logic like this 😆

src/routes/EResourceViewRoute.js
doytch: Another option that wouldn't require the stripes-connect could be to _not_ use `getRecords` for packageContents and write a custom `getPackageContentsRecords` for it. That function could check whether the records we already have are for an old filter because this component has access to a) the currently-selected package contents filter and b) the path used to fetch the current set of package contents (`resources.packageContents.url`). If that URL doesn't contain the current filter, we know the current set of records are stale and we should pass `data.packageContents` as `undefined` to our children.

src/routes/EResourceViewRoute.js
```diff
@@ -159,11 +167,14 @@ class EResourceViewRoute extends React.Component {
   }
 
   getRecords = (resource) => {
-    return get(this.props.resources, `${resource}.isPending`, true)
-      ?
-      undefined
-      :
-      get(this.props.resources, `${resource}.records`);
+    const records = get(this.props.resources, `${resource}.records`, []);
```
doytch: Doesn't this now change the behaviour of showing the loader for entitlements and entitlement options? It looks like we'll show the first page while we're fetching the full list. This makes me lean more [towards my second idea here](https://github.com/folio-org/ui-agreements/pull/404/files#diff-06b0e501c8006ad08ddf6c1656ad19acR206) since we could revert these changes.

src/routes/EResourceViewRoute.js
aditya-matukumalli: Yup, will change it

src/routes/EResourceViewRoute.js
aditya-matukumalli: Hmm, weird I dont think we even need `resultCount` here

PR #415, opened 2020-01-24
ERM-671: Added specific coverage handling for monographs

EthanFreestone:
PR to change the information displayed to the user in the `Coverage` column of `CoveredEResources` in the Agreement view and in the `PackageContents` in the EResource view, in the case of a monograph resource.


src/components/CoverageStatementsMonograph/CoverageStatementsMonograph.js
```diff
@@ -0,0 +1,79 @@
...
+
+  renderCoverage(pci) {
+    // Date can take the forms yyyy, yyyy-mm or yyyy-mm-dd, and is stored as a string.
+    const date = get(pci, 'pti.titleInstance.dateMonographPublished');
+    const volume = get(pci, 'pti.titleInstance.monographVolume');
+    const edition = get(pci, 'pti.titleInstance.monographEdition');
+
+    if (!date && !volume && !edition) {
+      return '*';
+    } else {
```
doytch: Note that when you return out of an `if` like you do on L50, you don't need this `else` statement and can break the contents of it back out into the parent scope.

src/components/BasketList/BasketList.js
```diff
@@ -22,6 +23,18 @@ class BasketList extends React.Component {
     selectedItems: PropTypes.object,
   }
 
+  coverageFormatter = (pci) => {
```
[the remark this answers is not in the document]
EthanFreestone: Have switched over to a new `<Coverage>` component, which uses `<SerialCoverage>` and `<MonographCoverage>`. In places where the old behaviour is required we import `SerialCoverage` and pass it `resource.coverage`, else we import the `Coverage` component and pass it the entire `resource`

src/components/AgreementLinesFieldArray/AgreementLineField.js
```diff
@@ -84,8 +84,7 @@ export default class AgreementLineField extends React.Component {
   }
 
   renderCoverage = (resource) => {
-    const coverage = get(resource, 'coverage');
-    return <CoverageStatement statements={coverage} />;
+    return <SerialCoverage statements={resource.coverage} />;
```
doytch: How do you know this is a serial? Couldn't the entitlement be for a book and thus a monograph?

EthanFreestone: Speaking to Gill, she mentioned that we wanted the behaviour in the edit screen to remain the same for the time being, so I used `SerialCoverage` for now, but this obviously would be very easy to change in future if needs be.

src/components/AgreementSections/CoveredEResourcesList.js
```diff
@@ -73,7 +73,7 @@ export default class CoveredEResourcesList extends React.Component {
     },
     platform: e => get(e._object, 'pti.platform.name', '-'),
     package: e => get(e._object, 'pkg.name', '-'),
-    coverage: e => <CoverageStatements statements={e.coverage} />,
+    coverage:  e => <Coverage pci={e._object} />,
```
doytch: Careful with the spacing
```suggestion
    coverage: e => <Coverage pci={e._object} />,
```

src/components/AgreementSections/LinesList.js
```diff
@@ -64,7 +64,7 @@ export default class LinesList extends React.Component {
     activeFrom: line => <div data-test-active-from>{this.renderDate(line.startDate)}</div>,
     activeTo: line => <div data-test-active-to>{this.renderDate(line.endDate)}</div>,
     count: line => <EResourceCount resource={getResourceFromEntitlement(line)} />,
-    coverage: line => <CoverageStatements statements={line.coverage} />,
+    coverage: line => <Coverage pci={line?.resource?._object} />,
```
doytch: Are you sure about this? It looks like you're now showing the base coverage and not custom coverage if it exists.

src/components/AgreementSections/CoveredEResourcesList.js
doytch: Are you sure about this? It looks like you're now showing the base coverage and not custom coverage if it exists.

test/bigtest/network/config.js
```diff
@@ -153,7 +153,9 @@ export default function config() {
   this.get('/erm/refdata/TitleInstance/type', () => {
     return [
       { 'id': '188389636d9ece46016d9ed0184b0027', 'value': 'journal', 'label': 'Journal' },
-      { 'id': '188389636d9ece46016d9ed018550028', 'value': 'book', 'label': 'Book' }
+      { 'id': '188389636d9ece46016d9ed018550028', 'value': 'book', 'label': 'Book' },
+      { 'id': '170227012020671001', 'value': 'serial', 'label': 'Serial' },
```
doytch: The output from that endpoint (against real backends) for me doesn't have Serial or Monograph entries. Am I missing something? Is that forthcoming? It doesn't look like it's being used.

EthanFreestone: Ah, I thought originally I'd need to add them there in order to use them in my objects, but realised afterwards I didn't have to, so these can go if they need to. I'm surprised that the endpoint doesn't return them for you, because they're 100% type options. I'll remove them for the time being since they're not in use in my test anyway.

src/components/AgreementLinesFieldArray/AgreementLineField.js
EthanFreestone: Comment added

src/components/Coverage/Coverage.js
```diff
@@ -0,0 +1,89 @@
...
+                monographVolume: PropTypes.string,
+                monographEdition: PropTypes.string,
+                name: PropTypes.string
+              })
+            })
+          })
+        })
+      })
+    }),
+    coveredEResource: PropTypes.shape({
```
doytch: Can you change this prop to just `resource` or `eresource`? A covered eresource is something very specific to agreements (it's an eresource we've been granted via agreement lines), and we're using the term for things like basket contents where it doesn't apply.

EthanFreestone: Have done :)

PR #416, opened 2020-01-24
Lazy load top level routing components in Agreements

aditya-matukumalli:
Added in the [Route based code splitting](https://reactjs.org/docs/code-splitting.html#route-based-code-splitting) which basically lazy loads the component to be rendered when we hit a specific route.

This gives us the advantage of not importing/including the component code associated with that route in the `bundle.js` thats served up when the app initially loads => technically smaller bundle size => faster load speeds.

Once we hit a specific route a chunk file `chunk.[hash].js` gets generated and served up which specifically contains the component code that needs to be rendered on that Route. For example: individual chunk files are generated for the Agreement view Route and Agreement Edit route which get served up only when accessing those specific routes. 

We make use of the combination of [dynamic imports](https://webpack.js.org/guides/code-splitting/#dynamic-imports) and [React lazy loading](https://reactjs.org/docs/code-splitting.html#reactlazy) to achieve this code-splitting. 

I am currently seeing a 1.7 MB difference in the bundle size on my local dev environment when using lazyLoading.


aditya-matukumalli:
> I really like this overall.
> 
> The only thing I keep turning over in my head is what the `Suspense` `fallback` should be. I experimented a bit with stripes-components' `LoadingView` since I don't love the tiny, little, margin-less `Spinner` at the top of the pane. Honestly I kinda wish we could just keep the previous view in place until the new chunk is picked up. Is that a possibility?

Yeah. Could just pass `fallback={null}` instead

PR #420, opened 2020-01-28
ERM-680 Show spinner in agreement lines when switching between agreem…

CalamityC:
…ents

add getAgreementEresourcesRecords function to AgreementViewRoute

### DIFF
--- a/src/routes/AgreementViewRoute.js
+++ b/src/routes/AgreementViewRoute.js
@@ -209,7 +209,6 @@ class AgreementViewRoute extends React.Component {
     return {
       ...agreement,
       contacts,
-      //  eresources: get(resources, 'agreementEresources.records'),
       eresources: this.getAgreementEresourcesRecords(),
       eresourcesCount: get(resources, 'agreementEresources.other.totalRecords'),
       lines: get(resources, 'agreementLines.records'),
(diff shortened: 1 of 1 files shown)


src/routes/AgreementViewRoute.js
```diff
@@ -236,6 +236,18 @@ class AgreementViewRoute extends React.Component {
     );
   }
 
+  getAgreementEresourcesRecords = () => {
+    const { resources, match } = this.props;
+    const agreementEresourcesUrl = get(resources, 'agreementEresources.url', '');
+    // If a new agreement is selected or if the filter has changed return undefined
+    return (agreementEresourcesUrl.indexOf(`${match.params.id}`) === -1 ||
+    agreementEresourcesUrl.indexOf(`resources/${resources.eresourcesFilterPath}`) === -1)
```
aditya-matukumalli: @CalamityC The spinner wouldn't show up here when you move from the any of the `current, future or dropped` filter to the `all` filter because `agreementEresourcesUrl.indexOf(resources/${resources.eresourcesFilterPath}` would not be equal to -1 as `eresourcesFilterPath` is an empty string in that case. To counter this Steve added an alias [here](https://github.com/folio-org/mod-agreements/blob/master/service/grails-app/controllers/org/olf/UrlMappings.groovy#L75) which should also be done [here](https://github.com/folio-org/mod-agreements/blob/master/service/grails-app/controllers/org/olf/UrlMappings.groovy#L16) for this logic to work

PR #424, opened 2020-01-30
Support Custom Properties in Agreements

aditya-matukumalli:
Added the following:
- Settings page to add/edit/delete the supplementary properties.
- Used CustomPropertiesConfigListFieldArray from erm-component to Add/Edit supplementary Properties to agreements
- Properties are listed under Supplementary properties accordion in view Agreements page
- Tests


src/components/views/Agreement.js
```diff
@@ -58,6 +59,7 @@ export default class Agreement extends React.Component {
   state = {
     showDuplicateAgreementModal: false,
     sections: {
+      supplementaryProperties: false,
```
doytch: The convention here is that the order is the visible order of the accordion sections.

src/components/views/Agreement.js
```diff
@@ -20,6 +20,7 @@ import { LoadingPane } from '@folio/stripes-erm-components';
 import DuplicateAgreementModal from '../DuplicateAgreementModal';
 
 import {
+  SupplementaryProperties,
```
doytch: The convention here is alphasorted

src/components/views/AgreementForm.js
```diff
@@ -20,13 +20,14 @@ import { AppIcon, TitleManager } from '@folio/stripes/core';
 import stripesFinalForm from '@folio/stripes/final-form';
 
 import {
+  FormSupplementaryProperties,
```
doytch: Same comment here and in `sections` below as in `Agreement.js`

src/components/views/AgreementForm.js
```diff
@@ -64,11 +65,12 @@ class AgreementForm extends React.Component {
   state = {
     addedLinesToAdd: false, // eslint-disable-line react/no-unused-state
     sections: {
+      formSupplementaryProperties: true,
       formInternalContacts: true,
       formLines: true,
       formLicenses: true,
       formOrganizations: true,
-      formSupplementaryInfo: true,
+      formSupplementarydocs: true,
```
doytch: camelCasing convention

PR #429, opened 2020-02-05
ERM-691 Current period note should display below the Cancellation dea…

CalamityC:
…dline

add row for currentPeriodNote in Info, add translation key

### DIFF
--- a/src/components/AgreementSections/Info.js
+++ b/src/components/AgreementSections/Info.js
@@ -30,7 +30,6 @@ export default class Info extends React.Component {
   render() {
     const { agreement } = this.props;
     const agreementIsClosed = get(agreement, 'agreementStatus.value') === statuses.CLOSED;
-    const currentPeriodNote = get(agreement.currentPeriod, 'note');
     return (
       <div data-test-agreement-info>
         <Row>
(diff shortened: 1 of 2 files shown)


src/components/AgreementSections/Info.js
```diff
@@ -76,6 +77,15 @@ export default class Info extends React.Component {
             </KeyValue>
           </Col>
         </Row>
+        <Row>
+          <Col xs={12}>
+            <KeyValue label={<FormattedMessage id="ui-agreements.agreements.currentPeriodNote" />}>
+              <div data-test-agreement-current-period-note style={{ whiteSpace: 'pre-wrap' }}>
+                {currentPeriodNote || '-'}
```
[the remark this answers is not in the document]
CalamityC: cool, thanks

CalamityC:
@doytch do you want me to change all occurrences of get in that file?

src/components/AgreementSections/Info.js
```diff
@@ -77,15 +76,11 @@ export default class Info extends React.Component {
...
-            <KeyValue label={<FormattedMessage id="ui-agreements.agreements.currentPeriodNote" />}>
-              <div data-test-agreement-current-period-note style={{ whiteSpace: 'pre-wrap' }}>
-                {currentPeriodNote || '-'}
-              </div>
-            </KeyValue>
-          </Col>
-        </Row>
+        <KeyValue label={<FormattedMessage id="ui-agreements.agreements.currentPeriodNote" />}>
+          <div data-test-agreement-current-period-note style={{ whiteSpace: 'pre-wrap' }}>
+            {agreement?.currentPeriod?.note ?? '-'}
```
doytch: I actually completely forgot about this, but reviewing another PR reminded me: we now have a [`NoValue` component in stripes-components](https://github.com/folio-org/stripes-components/blob/d6211488ebde5bf1f18d2dd04758cf580310e6ac/lib/NoValue/NoValue.js#L7).

The nice thing about that component is that it sets the `aria-label` so that a screenreader speaks out "no value set" rather than "dash".

doytch: So you could do something like `?? <NoValue />`

CalamityC: even cooler
i made the changes for all values, hope that's ok

PR #435, opened 2020-02-12
PR to allow vendor information to be viewed on the package info screen, if it exists

EthanFreestone:


src/components/EResourceSections/PackageInfo.js
```diff
@@ -60,7 +60,7 @@ export default class PackageInfo extends React.Component {
           <Col xs={4}>
             <KeyValue
               label="Provider"
-              value="-"
+              value={eresource?.vendor?.name || '-'}
```
doytch: Prefer `NoValue` component here.

EthanFreestone: Ah, yeah change made throughout

PR #487, opened 2020-04-14
ERM-805 Add Embargo to "E-resource in package" and "Agreement line" display

CalamityC:
* add Embargo component and use it in Coverage.js
 * add Embargo test to test/bigtest/tests/coverage-tests.js


src/components/Coverage/Coverage.js
```diff
@@ -78,11 +79,17 @@ export default class Coverage extends React.Component {
 
     if (isMonograph) {
```
aditya-matukumalli, line 80: Can you use declarative approach here using the ternary operator `?` So

`return isMonograph ? (...) : (...)`

src/components/Coverage/Coverage.js
```diff
@@ -78,11 +79,17 @@ export default class Coverage extends React.Component {
...
+          <MonographCoverage pci={dataToRender} />
+          <Embargo embargo={dataToRender.embargo} />
+        </div>
       );
     } else {
       return (
-        <SerialCoverage statements={dataToRender.coverage} />
+        <div>
+          <SerialCoverage statements={dataToRender.coverage} />
+          <Embargo embargo={dataToRender?.embargo ?? dataToRender?._object?.embargo ?? dataToRender?.resource?._object.embargo} />
```
aditya-matukumalli, line 91: Logic here is way too confusing. Can you just do this in the `if` blocks above? Like create a `let embargo` variable and populate them accordingly in the `if` blocks based on wether its a line, eresource or a pci

aditya-matukumalli, line 91: This line is kinda confusing. I think you could move this logic to the individual if blocks above like how we populate the `dataToRender` value to make it a bit more cleaner and consistent? Like define a `let embargo` variable and populate them based on wether its a pci, line or an eresource.

Also the `??` needs to be used only to assign a default value such as `undefined ?? {}` or `null || {}` but not as a substitue to the `||` operator

src/components/Embargo/Embargo.js
```diff
@@ -0,0 +1,68 @@
...
+          >
+            <FormattedMessage id="ui-agreements.embargo.start" />
+            <FormattedMessage id={unitId} values={{ count: number }} />
+          </Layout>
+        );
+      }
+      return null;
+    }
+
+
```
aditya-matukumalli, line 56: Remove whitespace

test/bigtest/interactors/embargo.js
```diff
@@ -0,0 +1,20 @@
+import {
+  interactor,
+  isPresent,
+  text,
+} from '@bigtest/interactor';
+
+@interactor class EmbargoInteractor {
+  static defaultScope = '[data-test-embargo]';
+
+  exists = isPresent('div');
```
aditya-matukumalli, line 10: Did you mean to say `isPresent('[data-test-embargo]')` here?

test/bigtest/tests/coverage-tests.js
```diff
@@ -53,27 +80,58 @@ const serial = {
...
-  resource: monographEResource
+  resource: monographEResource,
 };
 
 const serialLine = {
-  resource: serialEResource._object,
-  coverage: serial.coverage
+  resource: serial,
+  coverage: serial.coverage,
+  embargo: serial.embargo
```
aditya-matukumalli, line 134: Tests are failing becuase the code expects embargo to be on`line.resource._object.embargo` vs `line.resource.embargo` as you have it here. The shape of the data needs to be matched with that of the backend

src/components/Embargo/Embargo.js
```diff
@@ -0,0 +1,67 @@
...
+      if (embargo.movingWallStart) {
+        const unitId = `ui-agreements.${embargo.movingWallStart.unit}`;
+        const number = embargo.movingWallStart.length;
+        return (
+          <Layout
+            data-test-embargo-start
+            data-test-embargo-start-length={number}
+            data-test-embargo-start-unit={embargo.movingWallStart.unit}
+          >
+            <FormattedMessage id="ui-agreements.embargo.start" />
```
doytch, line 48: Can you verify this works in RTL languages like Arabic? 

If it works correctly, it should show up something like `60 days Moving wall start:` (flipping the order of the two FormattedMessages).

src/components/Coverage/Coverage.js
```diff
@@ -56,33 +57,43 @@ export default class Coverage extends React.Component {
     const { pci, line, eResource } = this.props;
     let isMonograph = false;
     let dataToRender;
+    let embargo;
```
doytch, line 60: Because of the various ways coverage is stored, this is unfortunately a pretty complicated component. Because of that the prop types are actually pretty valuable here in helping us understand what the expected shape is of the various props that can be passed in. So can you extend the prop types for this component to include where the embargo info is for the three ways this component can be called? And maybe include the `titleInstance.type.value` structure in it as well?

Because even after looking at the backend implementation and JIRA issues for this, I _still_ struggle to follow this and can't for sure tell whether everything's fine or something's being missed here 😆 (eg, sometimes we use `dataToRender.embargo` even after we save it off to `embargo`).

CalamityC: I've already had a conversation with Aditya about that. Because of my extreme laziness (home office is not good for me) I've tried to avoid extending the prop types. I should have known that I couldn't get away with it ...

src/components/Embargo/Embargo.js
CalamityC: No, it shows like in English `Moving wall start: 60 days`
How can I fix that?

src/components/Coverage/Coverage.js
doytch: Whenever we can make the code easier to understand it helps the rest of us who have the same problem right now. 🙃

src/components/Embargo/Embargo.js
doytch: [flexbox is a big help to us for RTL compatibility, particularly how it handles `flex-direction`](https://css-tricks.com/snippets/css/a-guide-to-flexbox/). So I believe all you may have to do is set that `Layout`'s className as [`flex`](https://github.com/folio-org/stripes-components/blob/8d75fda73bf8a6bcda3d29a22234ea41a728534e/lib/Layout/styles/flexbox.css#L5). 

If you run into alignment issues, it may be because FormattedMessage renders `span`s by default and flexbox works best when its flex items are `div`s. Therefore, setting the `tagName` prop on them to `"div"` should solve it.

test/bigtest/tests/coverage-tests.js
```diff
@@ -99,6 +122,58 @@ describe('Coverage tests', () => {
     it('correctly renders the volume', () => {
       expect(monographInteractor.volume).to.have.string(monograph.pti.titleInstance.monographVolume);
     });
+
```
doytch, line 125: Again, this is a complicated component with a bunch of if-case switching that's tough to follow. So [from what I see here](https://github.com/folio-org/ui-agreements/pull/487/files#diff-a21189892d9c449c206496ab1e32c75cR106), we've got six different paths. And each of those paths can either have coverage (of some form), or not have it. So at a minimum, I see a value for 12 different test cases.

And then separate from that, I'd have another unit test suite (`embargo-tests.js`) that just mounts the `Embargo` component and runs through all of it's permutations.

I _frequently_ think exhaustive testing is dumb and a waste of time, but this is definitely not one of those times. I see Coverage and Embargo as very risky components and worth having really bulletproof tests for, especially given how hard it is to experimentally validate everything on a live system.

CalamityC: I extended the existing coverage test with the embargo data. So if you recommend a separate embargo-test I would start with that, but leave the existing coverage test untouched as this issue is about embargo and not coverage.

doytch: So there are two prongs:
1. Tests are needed for Coverage for the six different paths because they change what's fed into Embargo. We cant test those paths via just Embargo.
2. Tests are needed for Embargo to confirm all its various permutations of embargoes. And yeah, those should be in a separate test that just mounts the Embargo component.

src/components/Embargo/Embargo.js
CalamityC: setting `Layout` className to flex didn't solve the issue

CalamityC: Yes, I tried that (although Mark mentioned that this should solve alignment issues) and it had no effect on the direction but broke the lines

doytch: [This is basically what we're trying to accomplish here.](https://codepen.io/doytch/pen/PoPNdJj) There may be an element or something in this that's either getting erroneously added or missed. 

Note that if you're looking for it, `dir="rtl"` is applied directly on the wrapping element in that example. However, in Folio, it'll be applied on the top-level `html` tag.
Screen Shot 2020-04-17 at 8 59 20 AM

It may be easiest to debug this by writing a simplified case of just `div`s and getting that to flip its direction. Then, either replace the container with Layout or the items with the FormattedMessage. Then, put it all together. Step by step.

aditya-matukumalli: image
 I see this on the front-end when I just pass `tagName="div"` to the `FormattedMessage` s that we render in the `renderEmbargoEnd` and `renderEmbargoStart` functions....with a `className="flex"` on the containing `Layout`s

CalamityC: Hm, thanks Aditya. I tried it and it didn't help. 
I tried it again today and it looks like in your example, I don't understand what's happened

test/bigtest/tests/coverage-tests.js
CalamityC: Just for clarification:
> 1. Tests are needed for Coverage for the six different paths because they change what's fed into Embargo. We cant test those paths via just Embargo.

I kept the existing 6 coverage tests (pci, eresource, line for monograph and serial each) and added only the embargo data. No changes on coverage data.

> 2. Tests are needed for Embargo to confirm all its various permutations of embargoes. And yeah, those should be in a separate test that just mounts the Embargo component.

I added an embargo test for 4 different embargo cases: embargo with Start and End, no embargo, embargo w/o end, embargo w/o start (although this should never occur)

Is that what you expected?

PR #491, opened 2020-04-16
ERM-827 Add support for "Alternative name" for agreements

CalamityC:
* change FormInfo and Info
* add alternateNames.name to searchKey in AgreementsRoute


src/components/AgreementSections/Info.js
```diff
@@ -98,6 +100,13 @@ export default class Info extends React.Component {
             </Col>
           </Row>
         }
+        {agreement.alternateNames !== undefined && agreement?.alternateNames?.length !== 0 &&
```
doytch, line 103: This is enough:
```suggestion
        {agreement.alternateNames?.length &&
```

PR #498, opened 2020-04-22
support empty array of alternateNames - refs ERM-827

CalamityC:
should not be merged before backend PR  https://github.com/folio-org/mod-agreements/pull/272


CalamityC:
when i change the backend to send an empty array and don't change the frontend code i see a 0 when no alternative names are set

PR #508, opened 2020-05-04
Create PCI display/edit screen

aditya-matukumalli:
- Added new PCI view/edit screens
- Added new EresourceEdit route (handles only pci resource currently)
- Added perm check to prevent non-pci edits
- Added new components to agreements. (TitleInfo, ParentPackageDetails, etc)
- Added Support for fixing incorrect coverage statements via pci edit page
- Tests 
    -  Added unit tests for the subcomponents to bypass the setup/teardown of BigTests
    -  Added a resources folder that contain all the data props required to inject into these components 
    - Edited TestForm helper form to accept initialValues prop


src/components/EResourceSections/SerialResourceInfo.js
```diff
@@ -0,0 +1,65 @@
...
+            </KeyValue>
+          </Col>
+          <Col xs={3}>
+            <KeyValue label={<FormattedMessage id="ui-agreements.eresources.materialType" />}>
+              <div data-test-title-instance-sub-type>{titleInstance.subType?.label ?? <NoValue />}</div>
+            </KeyValue>
+          </Col>
+        </Row>
+        <Row>
+          {renderIdentifier('ezb')}
```
doytch, line 57: `<Identifier>` possibility?

src/components/views/Agreement.js
```diff
@@ -129,7 +129,7 @@ export default class Agreement extends React.Component {
 
   getActionMenu = ({ onToggle }) => (
     <>
-      <IfPermission perm="ui-agreements.agreements.edit">
+      <IfPermission perm="ui-agreements.resources.edit">
```
doytch, line 132: ?

src/components/views/EResource.js
```diff
@@ -1,32 +1,64 @@
...
+      handlers,
+    } = this.props;
+
+    return (eresource?.class === resourceClasses.PCI) ? (
+      <IfPermission perm="ui-agreements.resources.edit">
+        <PaneMenu>
+          <FormattedMessage id="ui-agreements.eresources.edit">
+            {ariaLabel => (
+              <Button
+                aria-label={ariaLabel}
```
doytch, line 44: Don't set aria-label if it's just the same thing as the button label itself. It makes the screen reader read out the same thing twice.

src/components/views/PCI.js
```diff
@@ -0,0 +1,169 @@
...
+      }).isRequired,
+      relatedEntitlements: PropTypes.array,
+    }),
+    handlers: PropTypes.shape({
+      onClose: PropTypes.func.isRequired,
+    }),
+    isLoading: PropTypes.bool,
+  }
+
+  state = {
```
doytch, line 47: Let's not write new views with the old AccordionSet model that relies on us writing the section togglers. [Use `AccordionStatus` instead.](https://github.com/folio-org/stripes-components/blob/24665cc3e5e7e56811ea31466cb3c3d717bb4c73/lib/Accordion/readme.md#L42)

src/components/views/PCI.js
```diff
@@ -0,0 +1,169 @@
...
+  renderLoadingPane = () => {
+    return (
+      <Pane
+        defaultWidth="45%"
+        dismissible
+        id="pane-view-pci"
+        onClose={this.props.handlers.onClose}
+        paneTitle={<FormattedMessage id="ui-agreements.loading" />}
+      >
+        <Layout className="marginTop1">
```
doytch, line 88: Use the stripes-components [`LoadingPane` instead.](https://github.com/folio-org/ui-erm/blob/1d76c8608c31bc74729d7e287956e614511344f1/src/components/views/Agreement.js#L243)

src/components/views/PCIForm.js
```diff
@@ -0,0 +1,193 @@
...
+    pristine: PropTypes.bool,
+    submitting: PropTypes.bool,
+    values: PropTypes.object,
+  }
+
+  static defaultProps = {
+    initialValues: {},
+  }
+
+  state = {
```
doytch, line 44: Same as the view, use AccordionStatus instead of controlling state yourself.

src/components/views/PCIForm.js
```diff
@@ -0,0 +1,193 @@
...
+              id="close-pci-form-button"
+              onClick={this.props.handlers.onClose}
+            />
+          )}
+        </FormattedMessage>
+      </PaneMenu>
+    );
+  }
+
+  renderLoadingPane = () => {
```
doytch, line 128: This function is never called. If it's supposed to be, use `<LoadingView>` instead.

src/components/views/PCI.js
```diff
@@ -0,0 +1,169 @@
...
+            >
+              <FormattedMessage id="stripes-components.button.edit" />
+            </Button>
+          )}
+        </FormattedMessage>
+      </IfPermission>
+    </PaneMenu>
+  );
+
+  renderPCIInfo = (eresource) => <PCIInfo pci={eresource} />;
```
doytch, line 114: You're really overusing these render methods. There's no need to break them out of the main `render` function for these examples since it just forces us to scroll back and forth a bunch and pass variables along. `renderFoo` methods are useful when you have some logic you need to do in a computational scope to determine what to render. That's not happening here.

Additionally, what ends up happening is you lose track of code, like you did above with the `renderEditMenu` method. Similarly, in PCIForm, you lost track of the `renderLoadingPane` method.

src/routes/EResourceEditRoute.js
```diff
@@ -0,0 +1,98 @@
...
+
+  fetchIsPending = () => {
+    return Object.values(this.props.resources)
+      .filter(r => r && r.resource !== 'pci')
+      .some(r => r.isPending);
+  }
+
+  render() {
+    if (!this.state.hasPerms) return <NoPermissions />;
+    if (this.fetchIsPending()) return <LoadingPane onClose={this.handleClose} renderPaneset />;
```
doytch, line 84: Use stripes-components' `LoadingView` instead.

src/routes/EResourceViewRoute.js
```diff
@@ -211,18 +228,21 @@ class EResourceViewRoute extends React.Component {
...
           packageContentsCount: this.getPackageContentsRecordsCount(),
+          relatedEntitlements: this.getRecords('relatedEntitlements'),
         }}
         handlers={{
           ...handlers,
           onFilterPackageContents: this.handleFilterPackageContents,
           onNeedMorePackageContents: this.handleNeedMorePackageContents,
           onClose: this.handleClose,
+          onEdit: this.handleEdit,
+          onEresourceClick: this.handleEresourceClick,
```
doytch, line 245: Lets avoid introducing a third way of casing "eresource". 😆 Just use `*EResourceClick` here.

test/bigtest/interactors/pci-agreements.js
```diff
@@ -0,0 +1,24 @@
+import {
+  interactor,
+  text,
+} from '@bigtest/interactor';
+
+import MultiColumnListInteractor from '@folio/stripes-components/lib/MultiColumnList/tests/interactor';
+
+export default @interactor class PCIAgreementsInteractor {
+  agreementsList = new MultiColumnListInteractor('#pciagreements-agreements');
+  relatedAgreementsList = new MultiColumnListInteractor('#pcirelatedagreements-agreements');
```
doytch, line 10: `id`s should generally be `kebab-cased`. Sometimes that was awkward with accordion sections when we had to manually manage the open/close state so we used `camelCasing` but I don't think we need to do that anymore with AccordionStatus 🤔

test/bigtest/interactors/pci-edit.js
```diff
@@ -0,0 +1,29 @@
...
+}
+
+export default @interactor class PCIEditPaneInteractor {
+  isCoverageAccordionPresent = isPresent('#accordion-toggle-button-pciformcoverage');
+  isAddButtonPresent = isPresent('#edit-pci-add-coverage-button')
+  pciName = value('[data-test-pci-name]');
+  accessFrom = value('#pci-access-from');
+  accessUntil = value('#pci-access-end');
+
+  coverageCards = collection('#pci-form-coverages', coverageCardInteractor)
```
doytch, line 26: Does this work? Your tests only appear to ever index into the 0th element so it doesn't actually check if doing `coverageCards(n)` will indeed get the nth coverage.

test/bigtest/tests/pci-edit-test.js
```diff
@@ -0,0 +1,125 @@
...
+    it('correctly renders the add coverage button', () => {
+      expect(pciEditPaneInteractor.isAddButtonPresent).to.be.true;
+    });
+
+    describe('clicking the add coverage button', () => {
+      beforeEach(async () => {
+        await pciEditPaneInteractor.clickAddButton();
+      });
+
+      it('renderes a new coverage card', () => {
```
doytch, line 120: love a dash of Italian in my console logs 🍝

test/bigtest/tests/pci-edit-test.js
```diff
@@ -0,0 +1,125 @@
...
+
+    it('correctly renders the coverage end issue', () => {
+      expect(pciEditPaneInteractor.coverageCards(0).endIssue).to.equal(pci.coverage[0].endIssue);
+    });
+
+    it('correctly renders the add coverage button', () => {
+      expect(pciEditPaneInteractor.isAddButtonPresent).to.be.true;
+    });
+
+    describe('clicking the add coverage button', () => {
```
doytch, line 115: need tests to ensure that validations are hooked up and running correctly

test/bigtest/tests/pci-title-info-test.js
```diff
@@ -0,0 +1,111 @@
...
+        <Router context={{}}>
+          <TitleInfo
+            data={{ eresource: monographResource }}
+          />
+        </Router>
+      );
+    });
+
+    it('should render expected title', () => {
+      expect(interactor.titleInstanceName).to.equal(monographResource.pti?.titleInstance?.name);
```
doytch, line 31: _Never_ use an optional chain in an expected value. If the chain snaps and gives you `undefined`, you almost never want whatever you're `expect()`ing to be `undefined` as well. (In the rare case you do, you should use `.to.be.undefined()` or something.)

Sometimes with tests like this it almost makes more sense to just write out what you really want so it's very clear what should happen: `expect(interactor.titleInstaanceName).to.equal("Advances in Arithmetic")`

src/components/EResourceSections/AgreementsList.js
```diff
@@ -15,11 +15,14 @@ import CustomCoverageIcon from '../CustomCoverageIcon';
 import EResourceLink from '../EResourceLink';
 import EResourceType from '../EResourceType';
 import { getResourceFromEntitlement, urls } from '../utilities';
-import { resourceClasses } from '../../constants';
-
-const AgreementsList = ({ eresource, id, isRelatedEntitlement, resources }) => {
-  const isTitle = eresource.type !== undefined;
 
+const AgreementsList = (
```
doytch, line 19: Minor but trying to keep everything straight here since we now are gonna have like three or four components that render lists of agreements:
- Can we call this something like `EntitlementAgreementsList` or something? Just to underscore that you're actually passing in _entitlements_ and not agreements?
- On that point, can you call the prop `entitlements` instead of `resources`? And maybe use `e` instead of `res` in the arrow functions? Just to keep things clear when we look at this code again in a year.
- Not sure...but maybe at that point drag this out into its own folder at `src/components` rather than inside `EResourceSections`. The `Sections` folders should mostly be only filled with non-reusable components that make up a view.

src/components/EResourceSections/Identifier.js
```diff
@@ -0,0 +1,35 @@
...
+import PropTypes from 'prop-types';
+import { FormattedMessage } from 'react-intl';
+import {
+  Col,
+  KeyValue,
+} from '@folio/stripes/components';
+
+import { getResourceIdentifier } from '../utilities';
+
+const Identifier = (
```
doytch, line 11: Can we call this something like `EResourceIdentifier` instead so it's obvious what this component is? And move it up a level out of `EResourceSections`?

src/components/views/EResource.js
doytch: Sorry, I meant to just remove aria-label. It's obvious what the Edit button is going to do, so an aria attribute isn't necessary for clarification.

src/components/EResourceSections/AgreementsList.js
aditya-matukumalli: Done...Also added a boolean within the Agreements component wether to render the `relatedEntitlements` or not. Had to do this, if not the parent Accordion wasnt toggling as expected as `relatedEntitlements` call was bypassing the component containing  the `Accordion` .

src/components/views/PCIForm.js
aditya-matukumalli: Moved back to the old implementation as it seems like `AccordionStatus` doesnt go well with just a single `Accordion` child which is the case here.

doytch: If we only have a single Accordion then we don't need any of that state handling because we don't need an expand-all button, though. Can just have a single uncontrolled Accordion.

(Can you let @JohnC-80 know about the bug you're running into?)

test/bigtest/tests/pci-edit-test.js
aditya-matukumalli: Added the tests. Apparetly the browser had to be active while running the tests. (Specifically for fillAndBlur).

aditya-matukumalli:
@doytch I let John know about the AccordionStatus and it indeed was a bug which he would handle. Also Gill suggested to keep the ExpandAll button as is eventhough we just have a single accordion there. So, I left the logic in there as is.

aditya-matukumalli:
> * When viewing a title, is it intentional that title name be listed in a Headline _and_ in the Card's header? I didn't see a mockup for this scenario so not sure what Gill was envisioning.
> * There's permanently a Spinner in the `Agreements for this e-resource` section, even though all network requests have completed.
> * See example of both below:
> 
> Screen Shot 2020-05-06 at 10 50 49 AM
> 
> - When viewing a package, the `onRowClick` of the `E-resources in package` doesn't appear to be working. Only the entry in the Name column directs the user anywhere (since it's a `Link` [and it's only to the Title page, not to the PCI page]). I _thought_ the entire row was supposed to be clickable. If it is, it needs fixing. If it's _not_, then the MCL shouldn't be `interactive`. Double check with Gill? - I actually haven't been able to navigate to the PCI page at all. Can you give an example of how I'm supposed to get there?

Yeah you are right...That link in the header doesnt make sense as it points to the same view. I will check with Gill as to what needs to be done here?

Will fix the spinner issue

The PCI needs to be accessed from `options for acquiring this e-resource` from a `TitleInstance` view. The whole row is selectable there. I will add the interactive false for the MCL you have pointed out

PR #510, opened 2020-05-06
ERM-875 Use standard coverage display in "Agreements for this resourc…

CalamityC:
…e" and "Options for acquiring e-resource"

* use Coverage component instead of SerialCoverage component in EResourceSections/Agreements.js
* use eResource object instead of pci object in Coverage component in EResourceSections/AcquisitionOptions.js


CalamityC:
> However, you've got some merge conflicts here that need to be sorted out and I'm not 100% certain about the correct resolution of them. So since it's @aditya-matukumalli's PR that you're in conflict with, can you work with him to figure out what needs to be done ?

It was easy to solve, @aditya-matukumalli moved the code from  `EResourceSections/Agreements.js` to `EntitlementAgreementsList`, so I restored EResourceSections/Agreements.js from master, merged master into erm-875 and made my changes in EntitlementAgreementsList.

PR #511, opened 2020-05-06
ERM-867 Use titleInstance longName instead of name in eresource searc…

CalamityC:
…h results


src/components/views/EResources.js
```diff
@@ -227,6 +227,7 @@ const EResources = ({
                       isbn: e => getResourceIdentifier(e._object, 'isbn'),
                       eissn: e => getResourceIdentifier(e._object, 'eissn'),
                       pissn: e => getResourceIdentifier(e._object, 'pissn'),
+                      name: e => e._object?.longName ?? e.name,
```
doytch, line 230: Being pedantic here, but can you keep the order of the object keys so that it's in display order? Ie, `name` should be first, like it is in `visibleColumns` (and the other `column***` objects). Makes it easier to find what you're looking for 🤷

PR #513, opened 2020-05-08
ERM-870 ERM-871: Added Agreement Line view

doytch:
Putting this PR up so we can merge it and get intermediate feedback on MCL UX. It contains the agreement line view functionality but not editing.

- Added routes for viewing/editing agreement lines.
- Added view for viewing agreement lines per ERM-871
- Reorganized the Title/Package info/card components and updated their tests.
  - They're now at `TitleCard` and `PackageCard` instead of inside `EResourceSections`.
- Created a new `POLineCard` component that standardizes the display.
  - Used it in the agreement line view.


src/components/AgreementSections/Lines.js
```diff
@@ -16,6 +16,7 @@ export default class Lines extends React.Component {
     }).isRequired,
     eresourcesFilterPath: PropTypes.string,
     handlers: PropTypes.shape({
+      onViewAgreementLine: PropTypes.func.isRequired,
```
aditya-matukumalli, line 19: Nit-pick...alphasort..

PR #521, opened 2020-05-19
ERM-889 Add new agreement-agreement relationships

CalamityC:


src/components/RelatedAgreementsFieldArray/RelatedAgreementsFieldArray.js
```diff
@@ -32,7 +32,9 @@ class RelatedAgreementsFieldArray extends React.Component {
 
     agreementRelationshipTypes.forEach(type => {
       this.relationshipTypes.push(type.outward);
-      this.relationshipTypes.push(type.inward);
+      if (type.inward.value !== 'related_to_inward') {
```
doytch, line 35: It might be more obvious why we're doing this if we just key off of `type.type` (oof, bad naming by me 🤦), but can you leave a comment here explaining why we're excluding related-to?

PR #525, opened 2020-05-21
Async validate agreement name

aditya-matukumalli:
- Added asynchronous validattion to check for the uniqueness of the agreement name.
- Used withAsyncValidation HOC


src/components/AgreementSections/FormInfo.js
```diff
@@ -67,6 +73,13 @@ export default class FormInfo extends React.Component {
     return null;
   }
 
+  validateUniqueProperty = (value, _, meta) => {
+    const { checkAsyncValidation, values = {} } = this.props;
+    if (!value || value === values[meta.name]) return '';
```
doytch, line 78: What's the second part of the OR here doing? "If the currently entered value is equal to...the currently entered value"? Are `values` (which, sidenote, isn't that what the `_` arg is?) the values that already passed validation and are part of the form state?

src/components/AgreementSections/FormInfo.js
```diff
@@ -67,6 +73,13 @@ export default class FormInfo extends React.Component {
     return null;
   }
 
+  validateUniqueProperty = (value, _, meta) => {
```
doytch, line 76: Technically this isn't strictly to do with uniqueness right? You only know it's uniqueness bc you know that that's the only validation the backend has implemented. Maybe something more generic just talking about backend validation.

src/constants/validationEndPoint.js
```diff
@@ -0,0 +1,5 @@
+const validationEndPoint = {
```
doytch, line 1: Some of the names here need to be tweaked a bit. Either the file name, the constant, something. Because the constant's value is very specifically about an agreement. What happens when we add validation endpoints for other resources? Always plan ahead with naming conventions.

src/components/AgreementSections/FormInfo.js
aditya-matukumalli: Hmm `values`  actually needs to be `initialValues` here. So that the form lets us submit it even though we edit another field on it. Will change it

PR #530, opened 2020-05-28
ERM-774 Display information from siblingInstances in resources results

CalamityC:
* add getSiblingIdentifier

### DIFF
--- a/src/components/views/EResources.js
+++ b/src/components/views/EResources.js
@@ -227,7 +227,7 @@ const EResources = ({
                       type: e => <EResourceType resource={e} />,
                       isbn: e => getResourceIdentifier(e._object, 'isbn'),
                       eissn: e => getResourceIdentifier(e._object, 'eissn'),
-                      pissn: e => getSiblingIdentifier(e._object, 'issn'),
+                      pissn: e => getSiblingIdentifier(e._object, 'issn') ?? getResourceIdentifier(e._object, 'pissn'),
                     }}
                     id="list-eresources"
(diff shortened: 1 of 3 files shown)


src/components/views/EResources.js
```diff
@@ -227,7 +227,7 @@ const EResources = ({
                       type: e => <EResourceType resource={e} />,
                       isbn: e => getResourceIdentifier(e._object, 'isbn'),
                       eissn: e => getResourceIdentifier(e._object, 'eissn'),
-                      pissn: e => getResourceIdentifier(e._object, 'pissn'),
+                      pissn: e => getSiblingIdentifier(e._object, 'issn'),
```
doytch, line 230: This change makes it so that the the resource's `pissn` is _never_ used. I can't find any text stating that this is the intention on the JIRA; I assumed it meant "use the sibling `issn` _if_ the resource doesn't have a `pissn`". Can you confirm how it's intended to function w/ Owen or the model with Steve?

CalamityC: As I understood the implementation of **siblingIdentifiers** implies that print identifiers are no longer stored under the electronic resources **identifiers** but I can ask Owen / Steve

doytch: Huh. In that case, can you also ask Owen if we should change _all_ the places we use `pissn` and not just this one?
Screen Shot 2020-05-28 at 10 11 43 AM

CalamityC:
I'm closing this PR because of upcoming backend changes

PR #536, opened 2020-06-02
Add Suppress from Discovery field to PCI View/Edit

aditya-matukumalli:


src/components/EResourceSections/PCIFormInfo.js
```diff
@@ -49,6 +50,25 @@ export default class PCIFormInfo extends React.Component {
             />
           </Col>
         </Row>
+        <Row>
+          <Col xs={12}>
+            <Field
```
doytch, line 55: If this was c/ped from my code, [it doesn't need to be this complicated](https://github.com/folio-org/ui-agreements/pull/537).

aditya-matukumalli: Yeah, fixed it.

src/components/EResourceSections/PCIInfo.js
```diff
@@ -87,6 +87,15 @@ export default class PCIInfo extends React.Component {
             </KeyValue>
           </Col>
         </Row>
+        <Row>
```
doytch, line 90: Can you have this mirror the four-column layout in the edit page?

aditya-matukumalli: Done

PR #538, opened 2020-06-03
Change Eresource to be a functional component

aditya-matukumalli:
- Used the useIntl hook for the aria-label


src/components/views/EResource.js
```diff
@@ -10,48 +10,47 @@ import Title from './Title';
...
+  isLoading: PropTypes.bool,
+};
 
-  renderEditEresourcePaneMenu = () => {
-    const {
-      data: { eresource = {} },
-      handlers,
-    } = this.props;
+const EResource = ({
+  data: { eresource = {} },
```
doytch, line 32: Always take opportunities like this to clean things up. There's no need to destructure `eresource` and `data` individually. Find the places where you're using `data.eresource` and simplify them.

src/components/views/EResource.js
```diff
@@ -10,48 +10,47 @@ import Title from './Title';
...
+  data = {},
+  helperApp,
+  handlers,
+  isLoading
+}) => {
+  const intl = useIntl();
 
+  const renderEditEresourcePaneMenu = () => {
```
doytch, line 40: Again, use this opportunity to simplify things. You don't need a callback, you don't need a function. Can't you just put the ternary and JSX directly in the `lastMenu` prop?

src/components/views/EResource.js
aditya-matukumalli: Yeah, so we need both as we are passing the data prop down to the `EResourceViewComponent` component.

PR #541, opened 2020-06-08
ERM-774-776 Display information from siblingInstances in resource results and preview pane

CalamityC:
* add RelatedTitleInfo component and use it in TitleCardInfo
* change 'eISSN/ISSN' translation key
* add getSiblingIdentifier.js
* modify getResourceIdentifier.js
* add related-titles-test


src/components/utilities/getResourceIdentifier.js
```diff
@@ -1,23 +1,7 @@
...
-  if (Array.isArray(value)) {
-    return (
-      <Layout className="display-flex flex-direction-column">
-        {value.map((v, i) => <div key={i}>{v}</div>)}
-      </Layout>
-    );
-  }
+  const identifiers = eresource?.identifiers || [];
+  // find first identifiers value of namespace 'type'
+  const value = identifiers.find(i => i.identifier.ns.value === type)?.identifier?.value;
```
doytch, line 4: This appears to no longer have [the array-support that was here](https://github.com/folio-org/ui-agreements/pull/541/files?file-filters%5B%5D=.js&file-filters%5B%5D=.json#diff-0249252a51792ac88514adbaed7ad8b9L14-L20). Why is that no longer needed?

CalamityC: Because `find` is returning one element, the first. So how could we ever have an array there?

src/components/RelatedTitleInfo/RelatedTitleInfo.js
```diff
@@ -0,0 +1,78 @@
...
+    }),
+  })).isRequired,
+};
+
+const RelatedTitleInfo = ({
+  relatedTitles
+}) => (
+  relatedTitles.map((titleInstance, i) => (
+    <div key={`related-title-${i}`} data-test-related-titles>
+      <div className={css.separator} />
```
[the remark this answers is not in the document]
CalamityC: Sorry I don't understand what you mean with that.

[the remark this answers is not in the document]
CalamityC: > We shouldn't be rendering a separator for the last item in a list, this renders the separator unconditionally.
> 
Hm, but in this case we have to render the separator at the beginning of a related title section, we have to start with the separator because the title info always comes first and has no separator at its end
> Maybe adding that functionality _will_ give you a reason to keep this component as something that renders an entire array rather than my suggestion above.
yes I think so

src/components/RelatedTitleInfo/RelatedTitleInfo.js
```diff
@@ -14,64 +14,61 @@ import EResourceIdentifier from '../EResourceIdentifier';
...
-            <EResourceIdentifier titleInstance={titleInstance} type="doi" />
-          </>
-          :
-          <>
-            <EResourceIdentifier titleInstance={titleInstance} type="ezb" />
-            <EResourceIdentifier titleInstance={titleInstance} type="zdb" />
-            <EResourceIdentifier titleInstance={titleInstance} type="eissn" />
-            <EResourceIdentifier titleInstance={titleInstance} type="issn" />
-          </>
+  <div key={`related-title-${index}`} data-test-related-titles>
```
doytch, line 46: Setting the `key` isn't necessary here any more since you're setting it on the component itself. As a result, it appears that you no longer need to pass in `index` as a prop either.

PR #557, opened 2020-06-17
ERM-948 Enhance eHoldings display in dedicated Agreement Line view

CalamityC:
* add PackageCardExternal, TitleCardExternal for use in AgreementLineSections/Info.js
*  add test for agreement line view of external resources
* adjust EResourceCount to be able to show count of a parent package of an external resource


CalamityC:
> One thing, even though it's not called out in the JIRA, is how this will be used in the agreement line edit screen.

I completely forgot that an edit screen exists for an agreement line.
I changed  `FormInfo` of `AgreementLineSections` and the related test.
Additionally I inserted the `Title on platform URL`  in the view screen.

src/components/AgreementLineSections/Info.js
```diff
@@ -80,29 +83,45 @@ const Info = ({
...
+        { isExternal(line) ?
+          <TitleCardExternal title={resource} />
+          :
+          <TitleCard title={resource} />
+        }
+        <Headline size="large" tag="h3">
+          <FormattedMessage id="ui-agreements.eresources.parentPackageDetails" />
+        </Headline>
+        { isExternal(line) ?
+          <PackageCardExternal packageData={resource.reference_object.packageData} />
```
aditya-matukumalli, line 121: Should we add in a `?`  check here for `resource.reference_object?.packageData` here? or is it assumed that `reference_object` always exists on the `resource` object?

test/bigtest/tests/agreement-line-external-info-test.js
```diff
@@ -0,0 +1,232 @@
...
+          <Router context={{}}>
+            <Info line={externalResource} resource={externalResource} />
+          </Router>
+        );
+      });
+
+      // PARENT AGREEMENT
+      // resource = externalResource
+      it('should render parent agreement name', () => {
+        expect(interactor.parentAgreementName).to.equal(externalResource?.owner?.name);
```
aditya-matukumalli, line 33: Throughout these tests we are adding the `?` check on the `externalResource` and its nested objects as its expected value. If the chaining snaps that would return the expected value to be `undefined` and you almost never want whatever you're expect()ing to be undefined. 

So if possible you could just hardcode the actual expected value in there or instead just have it as `to.equal(externalResource.owner.name)` instead?

src/components/AgreementLineSections/Info.js
CalamityC: `reference_object`  should always exist but I added a `?` anyway

PR #609, opened 2020-09-15
ERM-780 Agreement lines: Enable MCL sorting

CalamityC:
add onSort function for name, activeFrom, activeTo to LinesList


aditya-matukumalli:
@CalamityC https://issues.folio.org/browse/ERM-780 asks for sorting on `Provider` and `PO Line ` columns as well. Any specific reason those columns have been left out?

CalamityC:
> @CalamityC https://issues.folio.org/browse/ERM-780 asks for sorting on `Provider` and `PO Line ` columns as well. Any specific reason those columns have been left out?

Yes, see Marks [comment](https://issues.folio.org/browse/ERM-780?focusedCommentId=74008&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-74008)

src/components/AgreementSections/LinesList.js
```diff
@@ -30,12 +30,23 @@ export default class LinesList extends React.Component {
...
+  };
+
   columnWidths = {
     name: 250,
     provider: 150,
     coverage: 250,
   }
 
+  sortMap = {
+    name: line => line.resource.name,
```
aditya-matukumalli, line 45: I think pulling in the `name` in the sortMap from `line.resource.name` will be an issue because the name might not always live under that object. For example: For a detached line we may not have a name and only have a description. This will probably break the UI. So firstly `line.resource.name` needs to be optional chained `line.resource?.name`  and then also all the different cases like `detached` line, `eholdings` line and `internal` line need to be tested for sorting.

aditya-matukumalli (changes requested):

src/components/AgreementSections/LinesList.js
CalamityC: I made the changes

aditya-matukumalli (approved):

PR #620, opened 2020-09-30
ERM-1141 Package view: UX improvements

CalamityC:
* enable sort in MCLs in package view
* add icon to EntitlementAgreementsList
* add columnWidth to PackageContents and EntitlementAgreementsList
* switch to 4 column layout in PackageInfo


src/components/EResourceSections/Agreements.js
```diff
@@ -32,16 +32,52 @@ export default class Agreements extends React.Component {
...
+    } = this.state;
+
+    // eslint-disable-next-line no-undef
+    const contentData = _.orderBy(entitlements,
+      [this.sortMap[sortOrder[0]], this.sortMap[sortOrder[1]]], sortDirection);
+
     return (
       <EntitlementAgreementsList
-        entitlements={entitlements}
+        contentData={contentData}
```
aditya-matukumalli, line 74: Any specific reason you changed the prop here Claudia?

src/components/EntitlementsAgreementsList/EntitlementAgreementsList.js
```diff
@@ -86,10 +108,13 @@ EntitlementAgreementsList.defaultProps = {
 };
 
 EntitlementAgreementsList.propTypes = {
-  entitlements: PropTypes.arrayOf(PropTypes.object),
+  contentData: PropTypes.arrayOf(PropTypes.object),
```
aditya-matukumalli, line 111: Same comment as above. We shouldnt be changing the API unless absolutely necessary. Also `contentData` is a prop that is speicifically used across FOLIO for the MCL. We should not be naming our app level prop the same as it might lead to confusion in the future.

aditya-matukumalli (changes requested):

src/components/EntitlementsAgreementsList/EntitlementAgreementsList.js
aditya-matukumalli: The tests also need to be fixed as a result of the prop change. Hopefully they will pass once you revert the change.

src/components/EResourceSections/Agreements.js
CalamityC: No, that (and the below) probably "happened" because I'm passing the `contentData`  from `Agreements` to `EntitlementAgreementsList`. I reverted the change.

src/components/EntitlementsAgreementsList/EntitlementAgreementsList.js
```diff
@@ -37,12 +43,25 @@ const EntitlementAgreementsList = (
...
+    name: e => {
+      const iconKey = e?.owner.agreementStatus?.value === statuses.CLOSED ? 'closedAgreement' : 'app';
+      return (
+        <AppIcon
+          app="agreements"
+          iconAlignment="baseline"
+          iconKey={iconKey}
+          size="small"
+        >
+          <div data-test-agreement-name><Link to={urls.agreementView(e?.owner?.id)}>{e?.owner.name}</Link></div>
```
aditya-matukumalli, line 61: `e?.owner.name` should actually be `e?.owner?.name`. But, this also tells that the name could be `undefined` here, so there needs to be a fallback. If not, `undefined` would be displayed on the UI

aditya-matukumalli: `data-test-agreement-name` is added here but I do not see any related tests?

src/components/EntitlementsAgreementsList/EntitlementAgreementsList.js
```diff
@@ -37,12 +43,25 @@ const EntitlementAgreementsList = (
...
   const columnWidths = {
+    name: { min: 200, max: 400 },
     startDate: 120,
     endDate: 120,
   };
 
   const formatter = {
-    name: e => <div data-test-agreement-name><Link to={urls.agreementView(e?.owner?.id)}>{e?.owner.name}</Link></div>,
+    name: e => {
+      const iconKey = e?.owner.agreementStatus?.value === statuses.CLOSED ? 'closedAgreement' : 'app';
```
aditya-matukumalli, line 53:
```suggestion
      const iconKey = e?.owner?.agreementStatus?.value === statuses.CLOSED ? 'closedAgreement' : 'app';
```

aditya-matukumalli (changes requested):

src/components/EntitlementsAgreementsList/EntitlementAgreementsList.js
CalamityC: I didn't touch this piece of code, I just wrapped it into `AppIcon`

aditya-matukumalli (approved):

PR #629, opened 2020-10-06
ERM-1121 pass alignment prop to Embargo component

CalamityC:


aditya-matukumalli (changes requested):
I feel it would be more appropriate if we have the `alignment` prop to be more generic such as `center`, `'left`, 'right` instead of actually passing in the className such as `centerContent`.

aditya-matukumalli (approved):

PR #660, opened 2020-11-04
ERM-1192 Load 100 values from user definable pick lists in agreements

CalamityC:
* add perPage param to agreement routes


aditya-matukumalli (changes requested):
Same [comment](https://github.com/folio-org/ui-licenses/pull/403#pullrequestreview-523438959)

aditya-matukumalli (approved):

PR #686, opened 2020-12-09
ERM-1228: Add POL/Inventory Instance details to Agreement Line display

EthanFreestone:
Added inventory link to POL Card when instanceId attached.
Amended tests/added translations

### DIFF
--- a/src/components/utilities/urls.js
+++ b/src/components/utilities/urls.js
@@ -34,7 +34,7 @@ const urls = {
 
   poLineView: id => `/orders/lines/view/${id}`,
 
-  viewInInventory: id => `/inventory/view/${id}`,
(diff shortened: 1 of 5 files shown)


src/components/utilities/urls.js
```diff
@@ -34,6 +34,8 @@ const urls = {
 
   poLineView: id => `/orders/lines/view/${id}`,
 
+  viewInInventory: id => `/inventory/view/${id}`,
```
aditya-matukumalli, line 37: Could probably change this to `viewInstance` instead?

aditya-matukumalli (changes requested):

src/components/utilities/urls.js
EthanFreestone: Done

PR #687, opened 2020-12-09
UIIN-1368: Rename instance-bulk to inventory-record-bulk

ViktorSoroka07:
## Purpose

Rename instance-bulk to inventory-record-bulk in the scope of the [UIIN-1368](https://issues.folio.org/browse/UIIN-1368).


ViktorSoroka07:
@aditya-matukumalli, please review.

PR #716, opened 2021-01-07
ERM-1460 Add test coverage for ui-agreements <MCLPaginationFields>

CalamityC:


src/settings/components/MCLPaginationFields/MCLPaginationFields.test.js
```diff
@@ -0,0 +1,40 @@
...
+import '@folio/stripes-erm-components/test/jest/__mock__';
+import { TestForm } from '@folio/stripes-erm-components/test/jest/helpers';
+import userEvent from '@testing-library/user-event';
+import MCLPaginationFields from './MCLPaginationFields';
+
+const onSubmit = jest.fn();
+const mclList = ['agreementLines', 'agreementEresources', 'entitlementOptions', 'packageContents', 'entitlements'];
+
+describe('MCLPaginationFields', () => {
+  test('renders mcl labels', () => {
```
aditya-matukumalli, line 12: We arent checking the rendering of the labels here right?

src/settings/components/MCLPaginationFields/MCLPaginationFields.test.js
```diff
@@ -0,0 +1,40 @@
...
+      </TestForm>
+
+    );
+
+    mclList.forEach((mcl) => {
+      expect(getByTestId(mcl)).toBeInTheDocument();
+    });
+  });
+
+  test('type value and render', () => {
```
aditya-matukumalli, line 25: We sould rather have a test that verifies the values on a form submit rather than in the actual TextField. So, you would have to spy on the `onSubmit` and verify if the field returns the expected value. Makes sense?

src/settings/routes/SupplementaryPropertiesConfigRoute.js
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { get } from 'lodash';
 import { stripesConnect } from '@folio/stripes/core';
-import SupplementaryPropertiesConfigForm from '../components';
+import SupplementaryPropertiesConfigForm from '../components/SupplementaryPropertiesConfigForm';
```
aditya-matukumalli, line 5: Please add a new `index.js` file at the top level within `src/settings/components` and export the `SupplementaryPropertiesConfigForm` from there. That way this could be imported as 
```suggestion
import SupplementaryPropertiesConfigForm from '../components';
```

test/bigtest/tests/suppress-from-discovery-tests.js
```diff
@@ -6,7 +6,7 @@ import { Form } from 'react-final-form';
 
 import { mountWithContext } from '../helpers/mountWithContext';
 
-import { SuppressFromDiscoveryFields } from '../../../src/settings/components';
+import SuppressFromDiscoveryFields from '../../../src/settings/components/SuppressFromDiscoveryFields';
```
aditya-matukumalli, line 9: Same comment as above

aditya-matukumalli (changes requested):

src/settings/components/MCLPaginationFields/MCLPaginationFields.test.js
CalamityC: How do we spy in react testing library? I didn't find an example yet.

src/settings/routes/SupplementaryPropertiesConfigRoute.js
CalamityC: Hm. I get it working with 
`import { SupplementaryPropertiesConfigForm } from '../components';`
as it's a named import?

src/settings/components/MCLPaginationFields/MCLPaginationFields.test.js
aditya-matukumalli: `jest.fn()` can be used to mock the callback.
So you could do 
`onSubmit = jest.fn()` and in your tests, you could enter a text in the TextField and check if `onSubmit` has been called with the expected values using the `toHaveBeenCalledWith` helper.

https://jestjs.io/docs/en/mock-functions#custom-matchers

src/settings/routes/SupplementaryPropertiesConfigRoute.js
aditya-matukumalli: Sounds good

PR #739, opened 2021-01-25
ERM-1532: Search by Agreement start / end date

EthanFreestone:
Added DateFilter components to AgreementFilters, making use of new props


src/components/AgreementFilters/AgreementFilters.js
```diff
@@ -235,12 +238,32 @@ export default class AgreementFilters extends React.Component {
...
+      activeFilters={this.props.activeFilters}
+      filterHandlers={this.props.filterHandlers}
+      name="startDate"
+    />;
+  }
+
+  renderEndDateFilter = () => {
+    return <DateFilter
+      activeFilters={this.props.activeFilters}
+      allowNull
```
aditya-matukumalli, line 252: Lets name this `showNoEndDateCheckbox`

src/components/AgreementFilters/AgreementFilters.js
```diff
@@ -235,12 +238,32 @@ export default class AgreementFilters extends React.Component {
...
+      filterHandlers={this.props.filterHandlers}
+      name="endDate"
+      resourceName={this.props.intl.formatMessage({ id: 'ui-agreements.agreements.lowerCase' })}
```
aditya-matukumalli, line 255: Cant we use `FormattedMessage` here?

aditya-matukumalli (changes requested):

src/components/AgreementFilters/AgreementFilters.js
EthanFreestone: I don't think so. I need to pass a string rather than an object. FormattedMessage returns a React node

src/components/AgreementFilters/AgreementFilters.js
EthanFreestone: But this could theoretically apply to either start or end date, for instance in Licenses. What about `showNoDateCheckbox`?

src/components/AgreementFilters/AgreementFilters.js
EthanFreestone: I'd rather use the hook than the messier HOC, but I didn't want to convert the entire filter component to functional

EthanFreestone: The easiest alternative I suppose would be a prop something like `noDateCheckboxLabel`, where we pass the entire label translation, removing that from the stripes-erm-components component

EthanFreestone: But that does seem overkill compared to passing "agreements" and "licenses" as strings to a ready made translation in the DateFilter comp as values

src/components/AgreementFilters/AgreementFilters.js
aditya-matukumalli: Thinking through this again, I feel `hideNoEndDateCheckbox` would be better here, so that we can default the shared component to show the checkbox and hide it conditionaly. Also, please add a default prop value of `false` to `hideNoEndDateCheckbox`

src/components/AgreementFilters/AgreementFilters.js
aditya-matukumalli: Lets make use of this opportunity to tweak that component to a functional component and bring in the `useIntl` hook

EthanFreestone: Done

src/components/AgreementFilters/AgreementFilters.js
EthanFreestone: Have done both

src/components/AgreementFilters/AgreementFilters.js
```diff
@@ -14,45 +20,33 @@ const FILTERS = [
...
       }
     });
 
-    if ((props.data?.tagsValues?.length ?? 0) !== state.tags.length) {
-      newState.tags = props.data.tagsValues.map(({ label }) => ({ value: label, label }));
+    if ((data?.tagsValues?.length ?? 0) !== filterState.tags?.length) {
+      newState.tags = data.tagsValues.map(({ label }) => ({ value: label, label }));
     }
+    if (Object.keys(newState).length) {
+      setFilterState(newState);
```
aditya-matukumalli, line 45: This is causing an infinite loop right now, as the useEffect hook depends on filterState which is getting updated everytime you called `setFilterState` here

PR #743, opened 2021-01-27
ERM-1281: Add test coverage for ui-agreements <NoPermissions>

peterboehm:


test/helpers/translationsProperties.js
```diff
@@ -1,10 +1,14 @@
 import translations from '../../translations/ui-agreements/en';
+import stripesSmartComponentsTranslations from '@folio/stripes-smart-components/translations/stripes-smart-components/en';
```
aditya-matukumalli, line 2: Hmm, I see we would have to bring in `stripes-smart-components` as a devDependency (thats the lint warning you are facing) here just for the sake of translations in tests. Lets avoid that. Please revert this PR to the previous commit and we will go ahead with that version

peterboehm:
Reverted the changes. 
Jenkins didn't say anything, but seems to pass: https://jenkins-aws.indexdata.com/job/folio-org/job/ui-agreements/view/change-requests/job/PR-743/

PR #748, opened 2021-01-29
ERM-1531, ERM-1534

EthanFreestone:
- Renamed OtherPeriods accordion to AllPeriods and removed non-current filter
- Refactored Info to functional comp
- Created new subcomponent InfoPeriods with current, previous and next tabs

IMPORTANT
Must be merged along with backend PR


src/components/AgreementSections/InfoPeriods.js
```diff
@@ -0,0 +1,126 @@
...
+  Col,
+  FormattedUTCDate,
+  Headline,
+  KeyValue,
+  Layout,
+  NoValue,
+  Row,
+} from '@folio/stripes/components';
+
+export default function InfoPeriods({ periods }) {
```
aditya-matukumalli, line 16: arrow fn

src/components/AgreementSections/InfoPeriods.js
```diff
@@ -0,0 +1,126 @@
...
+  /*
+   * This could be handled by callback, but since we need to know for display whether next/prev/current exist
+   * it seems more prudent to just fetch all three at once and then do our logic at this stage
+   */
+  const displayPeriodOptions = {
+    currentPeriod: periods?.find(p => p.periodStatus === 'current'),
+    nextPeriod: periods?.find(p => p.periodStatus === 'next'),
+    previousPeriod: periods?.find(p => p.periodStatus === 'previous')
+  };
```
aditya-matukumalli, line 25: How does an example `periods` array look like? You are basically trying to convert an array to an object (Map) here. I would suggest we use `array.reduce` for this purpose instead of looping through each period here

src/components/AgreementSections/InfoPeriods.js
```diff
@@ -0,0 +1,126 @@
...
+   * Set up which period is currently selected. Heirachy is Current > Next > Previous
+   * The empty array as a second argument to useEffect should mean this code only runs once to prevent infinite loops
+   */
+  const [selectedPeriod, setSelectedPeriod] = useState('current');
+  useEffect(() => {
+    if (displayPeriodOptions.currentPeriod === undefined) {
+      if (displayPeriodOptions.nextPeriod !== undefined) {
+        setSelectedPeriod('next');
+      } else {
+        setSelectedPeriod('previous');
```
aditya-matukumalli, line 37: Can we avoid working with `undefined`s? and instead default the values generated in displayPeriodOptions to maybe `{}` and work with them assuming each individual period there is an object?

aditya-matukumalli (changes requested):

aditya-matukumalli:
Also, now that the `EresourceSections` RTL tests are up, this may be a good opportunity to write an RTL tests for this component

src/components/AgreementSections/AllPeriods.js
```diff
@@ -10,68 +9,61 @@ import {
...
-  render() {
-    const {
-      agreement: { currentPeriod = {}, periods = [] },
-      id,
-    } = this.props;
-
-    const otherPeriods = periods.filter(period => period.id !== get(currentPeriod, 'id'));
+  ];
 
+  if (periods.length > 1) {
```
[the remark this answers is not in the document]
EthanFreestone: Yes, Owen wanted this to not display if we only have a single period, since that will show up in the tab part at the top.

src/components/AgreementSections/AllPeriods.js
```diff
@@ -10,68 +9,61 @@ import {
...
-      })),
-      currentPeriod: PropTypes.shape({
-        id: PropTypes.string,
-      }),
-    }).isRequired,
-    id: PropTypes.string,
-  }
-
-  columnMapping = {
+export default function AllPeriods({ agreement: { periods = [] }, id }) {
```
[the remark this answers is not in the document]
EthanFreestone: Sure, I forgot you could export them like that

src/components/AgreementSections/AllPeriods.js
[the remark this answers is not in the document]
EthanFreestone: Change made

src/components/AgreementSections/Info.js
```diff
@@ -12,107 +11,96 @@ import {
...
+      <Row>
+        <Col xs={12}>
+          <KeyValue label={<FormattedMessage id="ui-agreements.agreements.agreementDescription" />}>
+            <div data-test-agreement-description style={{ whiteSpace: 'pre-wrap' }}>
+              {agreement.description || <NoValue />}
+            </div>
+          </KeyValue>
+        </Col>
+      </Row>
+      {agreement?.alternateNames?.length !== 0 &&
```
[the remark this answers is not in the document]
EthanFreestone: Change made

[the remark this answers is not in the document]
EthanFreestone: Although I believe that was existing code I've ported over

src/components/AgreementSections/InfoPeriods.js
```diff
@@ -0,0 +1,126 @@
...
+   */
+  const displayPeriodOptions = {
+    currentPeriod: periods?.find(p => p.periodStatus === 'current'),
+    nextPeriod: periods?.find(p => p.periodStatus === 'next'),
+    previousPeriod: periods?.find(p => p.periodStatus === 'previous')
+  };
+
+  /*
+   * Set up which period is currently selected. Heirachy is Current > Next > Previous
+   * The empty array as a second argument to useEffect should mean this code only runs once to prevent infinite loops
```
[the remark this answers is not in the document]
EthanFreestone: Sure, comment tweaked

src/components/AgreementSections/InfoPeriods.js
EthanFreestone: Hmmm, no. I'm doing no conversion, simply extracting the period p where p.periodStatus is equal to a certian value. This is what find is for, no?

array.reduce is for creating a value based on all the elements in the array, which I am not doing.


As per the proptypes:
```
periods: PropTypes.arrayOf(PropTypes.shape({
    cancellationDeadline: PropTypes.string,
    endDate: PropTypes.string,
    periodStatus: PropTypes.string,
    startDate: PropTypes.string.isRequired,
  })),
```
So we have an array of objects, each of which will have a periodStatus (Could be null).

That find command simply finds the one with a certain periodStatus, or returns undefined otherwise

src/components/AgreementSections/InfoPeriods.js
EthanFreestone: So use something like Object.keys({}).isEmpty to check for blank..? I can do but isn't that just harder to understand. Why not work with undefined?

src/components/AgreementSections/Info.js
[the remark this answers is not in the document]
EthanFreestone: Turns out this was necessary. 
```
agreement?.alternateNames?.length &&
```

displayed a 0 for the length value, rather than using it as a truthy check. Reverted to previous code

src/components/AgreementSections/InfoPeriods.js
EthanFreestone: Change made anyway :)

src/components/AgreementSections/InfoPeriods.js
```diff
@@ -0,0 +1,131 @@
...
+  Row,
+} from '@folio/stripes/components';
+
+const InfoPeriods = ({ periods }) => {
+  /*
+   * This could be handled by callback, but since we need to know for display whether next/prev/current exist
+   * it seems more prudent to just fetch all three at once and then do our logic at this stage
+   */
+
+  const displayPeriodOptions = periods?.reduce((map, period) => {
```
aditya-matukumalli, line 22: If we default the `periods` to be an empty array, we dont need the `?` and we can get rid if the default `{}` as well

src/components/AgreementSections/InfoPeriods.js
```diff
@@ -0,0 +1,131 @@
...
+   * This is equivalent to the old "componentDidMount"
+   */
+  const [selectedPeriod, setSelectedPeriod] = useState('current');
+
+  useEffect(() => {
+    if (!displayPeriodOptions.current) {
+      if (displayPeriodOptions.next) {
+        setSelectedPeriod('next');
+      } else {
+        setSelectedPeriod('previous');
```
aditya-matukumalli, line 40: Can we refactor this to be
```
setSelectedPeriod(prevState => {
    if (prevState.current) {
      return displayPeriodOptions.next ? 'next' : 'previous';
    }
})

**Review by aditya-matukumalli — CHANGES_REQUESTED** [review]

File: src/components/AgreementSections/InfoPeriods.js (hunk 9 above)
Comment (EthanFreestone) [reply src/components/AgreementSections/InfoPeriods.js:40]: No, because the "selectedPeriod" is simply a string, so prevState will not have a .current

File: src/components/AgreementSections/InfoPeriods.js (hunk 8 above)
Comment (EthanFreestone) [reply src/components/AgreementSections/InfoPeriods.js:22]: Sure

File: src/components/AgreementSections/InfoPeriods.js (hunk 9 above)
Comment (EthanFreestone) [reply src/components/AgreementSections/InfoPeriods.js:40]: I can however make it: 
```
if (!displayPeriodOptions.current) {
      setSelectedPeriod(() => {
        return displayPeriodOptions.next ? 'next' : 'previous';
      })
    }
```

It needs the external if, if that is inside the setSelectedPeriod then when displayPeriodOptions does indeed contain `current`, the setSelectedPeriod returns undefined. I don't particularly want to have to return the value it's already set to if nothing needs to change, so I wrapped the whole setSelectedPeriod

aditya-matukumalli (approved):

PR #759, opened 2021-02-10
ERM-1537 Hide accordions in Agreement view where there is no content

CalamityC:
* render certain accordions conditionally in agreement detail view
*  edit changelog


EthanFreestone (commented):
Looks good to me in general, however I would suggest writing a RTL test which tests whether the accordions exist or not for each use case, with/without data. Should be quite simple, as they'd be mostly repetitions of the same test for different accordions

EthanFreestone (approved):
These look good

PR #761, opened 2021-02-22
ERM-1544: Add tag search to eresources search and sort

peterboehm:


src/components/EResourceFilters/EResourceFilters.js
```diff
@@ -134,13 +134,40 @@ export default class EResourceFilters extends React.Component {
     );
   }
 
+  renderTagsFilter = () => {
+    const dataOptions = this.props.data.tagsValues.map(({ label }) => ({ value: label, label }));
```
EthanFreestone, line 138: Recently I refactored AgreementFilters to a functional component instead of a class component, and it's a lot cleaner.
https://github.com/folio-org/ui-agreements/blob/master/src/components/AgreementFilters/AgreementFilters.js

However, if this is outside your range at this point, perhaps double check against what it looked like before (see below), because I don't think we should be grabbing the data options like this.
https://github.com/folio-org/ui-agreements/blob/7aa290f7e9b31305211e90bbeb0b7fab3ad76ea0/src/components/AgreementFilters/AgreementFilters.js#L204-L227

EthanFreestone (changes requested):
Mostly looks good.

src/components/EResourceFilters/EResourceFilters.js
peterboehm: I derived the code from here:
https://github.com/folio-org/ui-agreements/blob/7aa290f7e9b31305211e90bbeb0b7fab3ad76ea0/src/components/AgreementFilters/AgreementFilters.js#L35-L52
For Agreements and Licenses, the setting of the dataOptions is done centrally for all filters using states by the code above. For eresources, this was not implemented. 
Do you suggest to add an if clause to account for empty values or what other aspect are you referring to?

EthanFreestone: Hi Peter. The rest of the file gets its dataOptions and inserts them into state, see https://github.com/folio-org/ui-agreements/blob/6148a5969ada22079b230c997b307c4b4ec14a5a/src/components/EResourceFilters/EResourceFilters.js#L94
and https://github.com/folio-org/ui-agreements/blob/6148a5969ada22079b230c997b307c4b4ec14a5a/src/components/EResourceFilters/EResourceFilters.js#L33-L46

I'd prefer it if we kept that pattern, rather than having the renderTag filter refetch its dataOptions on every render

EthanFreestone: I think this may be quite simple, by editing the FILTERS static object and checking the getDerivedStateFromProps is doing as expected.

src/components/EResourceFilters/EResourceFilters.js
```diff
@@ -1,52 +1,48 @@
...
-    return null;
-  }
+    if (Object.keys(newState).length) {
+      setFilterState(prevState => ({ ...prevState, ...newState }));
+    }
+  }, [data, filterState]);
 
-  renderIsPackageFilter = () => {
-    const { activeFilters } = this.props;
+  const renderIsPackageFilter = () => {
```
EthanFreestone, line 45: This looks functionally the same as renderCheckboxFilter, and is leading to a code smell. Could you try to get the package filters to use the generalised code please?

EthanFreestone (changes requested):
Looks mostly good, I'd make the package filters use renderCheckboxFilter though if possible

PR #791, opened 2021-04-12
ERM-1547 Related agreements: Improve error management when the user attempts to link an agreement to itself

CalamityC:
* convert RelatedAgreementField to functional component
* display error message when an agreement is linked to itself
* fix message banner layout
* rename state variable and function and fix setting it back to false
* remove validateSelfLinking function from RelatedAgreementsFieldArray


src/components/RelatedAgreementsFieldArray/RelatedAgreementField.js
```diff
@@ -19,71 +20,86 @@ import { AppIcon, Pluggable } from '@folio/stripes/core';
...
+  }).isRequired,
+  meta: PropTypes.shape({
+    error: PropTypes.node,
+    touched: PropTypes.bool,
+  }).isRequired,
+  onAgreementSelected: PropTypes.func.isRequired,
+  parentAgreementId: PropTypes.string,
+  parentAgreementName: PropTypes.string,
+};
+export default function RelatedAgreementField({
```
EthanFreestone, line 47: Could we swap this to the same pattern we use elsewhere?

```
const RelatedAgreementField = ({ destructuredProps }) => {
 ...
}
export default RelatedAgreementField
```

src/components/RelatedAgreementsFieldArray/RelatedAgreementField.js
```diff
@@ -19,71 +20,86 @@ import { AppIcon, Pluggable } from '@folio/stripes/core';
...
-  }
+  let triggerButton = useRef(null);
 
-  componentDidMount() {
-    if (!get(this.props, 'input.value.id') && get(this.triggerButton, 'current')) {
-      this.triggerButton.current.focus();
+  const { change } = useForm();
+
+  useEffect(() => {
+    if (!input.value.id && triggerButton.current) {
```
EthanFreestone, line 63: `get` comes with null safety, can we swap this to `input?.value?.id` please?

EthanFreestone: Actually I think `input.value?.id` may be sufficient

src/components/RelatedAgreementsFieldArray/RelatedAgreementField.js
```diff
@@ -19,71 +20,86 @@ import { AppIcon, Pluggable } from '@folio/stripes/core';
...
+      triggerButton.current.focus();
     }
-  }
+  });
```
EthanFreestone, line 66: If this needs to happen once, on mount, then please add an empty dependency array `[]`. If it needs to happen any time input/ triggerButton changes, then you need to add the dependencies `[input, triggerButton]`

src/components/RelatedAgreementsFieldArray/RelatedAgreementField.js
```diff
@@ -19,71 +20,86 @@ import { AppIcon, Pluggable } from '@folio/stripes/core';
...
+export default function RelatedAgreementField({
+  agreement,
+  id,
+  input,
+  meta,
+  onAgreementSelected,
+  parentAgreementId,
+  parentAgreementName,
+}) {
+  const [selfLinked, setSelfLinked] = useState(false);
```
EthanFreestone, line 56: Can we please change this to `selfLinkedWarning` or something? It is supposed to stick around even after the field has been reset to undefined, and so is no longer selfLinked.

src/components/RelatedAgreementsFieldArray/RelatedAgreementField.js
```diff
@@ -19,71 +20,86 @@ import { AppIcon, Pluggable } from '@folio/stripes/core';
...
       <Pluggable
         dataKey={id}
         onAgreementSelected={onAgreementSelected}
-        renderTrigger={(props) => {
-          this.triggerButton = props.buttonRef;
+        renderTrigger={(pluggableRenderProps) => {
+          triggerButton = pluggableRenderProps.buttonRef;
 
-          const agreementName = get(this.props.agreement, 'name');
+          const agreementName = agreement.name;
```
EthanFreestone, line 86: `agreement?.name` in case called without an agreement

src/components/RelatedAgreementsFieldArray/RelatedAgreementField.js
```diff
@@ -144,53 +158,63 @@ export default class RelatedAgreementField extends React.Component {
...
-          <FormattedMessage id="ui-agreements.relatedAgreements.noneLinked" />
-        </strong>
-      </Layout>
-      <Layout className="textCentered">
-        <FormattedMessage id="ui-agreements.relatedAgreements.linkToStart" />
-      </Layout>
-    </div>
-  )
+  const renderEmpty = () => {
+    return (
```
EthanFreestone, line 164: With functional components you can exchange

```
const functionName = () => {
  return (
    "hello  world"
  );
}
```
with

```
const functionName = () => (
  "hello world"
);
```

Ie if your function only consists of a return block, you can return straight away.

EthanFreestone: I think you can change the above function `renderAgreement` as well

EthanFreestone (changes requested):
A few minor tweaks, but overall looks good :)

src/components/RelatedAgreementsFieldArray/RelatedAgreementField.js
```diff
@@ -19,71 +20,86 @@ import { AppIcon, Pluggable } from '@folio/stripes/core';
...
-    const {
-      id,
-      input: { name },
-      onAgreementSelected,
-    } = this.props;
+  useEffect(() => {
+    if (parentAgreementId === input.value?.id && !selfLinkedWarning) {
+      change(input.name, undefined);
+      setSelfLinkedWarning(true);
+    } else if (input.value && parentAgreementId !== input.value?.id && selfLinkedWarning) {
```
EthanFreestone, line 72: I actually don't think you need the else here, since the checks are complete opposites of them. Two separate `if` blocks will suffice

src/components/RelatedAgreementsFieldArray/RelatedAgreementField.js
```diff
@@ -19,71 +20,86 @@ import { AppIcon, Pluggable } from '@folio/stripes/core';
...
+      setSelfLinkedWarning(false);
+    }
+  }, [change, input, parentAgreementId, selfLinkedWarning, setSelfLinkedWarning]);
 
+  const renderLinkAgreementButton = value => {
+    const name = input.name;
```
EthanFreestone, line 78: This probably isn't necessary, you can just replace name with input.name further down

EthanFreestone (changes requested):

EthanFreestone (approved):

PR #811, opened 2021-06-02
ERM-1609: Hide supplementary properties accordion if zero populated terms and no primary terms

peterboehm:


src/components/views/Agreement.js
```diff
@@ -200,6 +200,16 @@ class Agreement extends React.Component {
     };
   }
 
+  showTermsAccordion = () => {
```
EthanFreestone, line 203: I think this method should be called showSupplementaryPropertiesAccordion

src/components/views/Agreement.js
```diff
@@ -200,6 +200,16 @@ class Agreement extends React.Component {
...
+    const { data } = this.props;
+    const primaryprops = data.supplementaryProperties?.filter(p => p.primary === true).length;
+    let custprops = false;
```
EthanFreestone, line 206: We should probably initialise this to 0 to stay consistent with this variable holding a length later on.

EthanFreestone (changes requested):

PR #816, opened 2021-06-15
ERM-1691: Sort on Agreement Status should sort by label

CalamityC:


CalamityC:
PR closed because trying another approach

PR #823, opened 2021-06-23
ERM-1750 General License note not displayed in Agreements view

CalamityC:
* add conditional rendering of licenseNote in agreements view info section

### DIFF
--- a/src/components/AgreementSections/Info.js
+++ b/src/components/AgreementSections/Info.js
@@ -62,6 +62,19 @@ const Info = ({ agreement }) => {
           </KeyValue>
         </Col>
       </Row>
+      {agreement?.licenseNote ?
+        <Row>
+          <Col xs={12}>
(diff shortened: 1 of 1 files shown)


src/components/AgreementSections/Info.js
```diff
@@ -62,6 +62,19 @@ const Info = ({ agreement }) => {
           </KeyValue>
         </Col>
       </Row>
+      {agreement?.licenseNote ?
```
EthanFreestone, line 65: I think elsewhere we use the pattern
```
agreement?.licenseNote ?? <content />
```

Other than that, seems good

EthanFreestone (changes requested):

src/components/AgreementSections/Info.js
CalamityC: It's 
`agreement?.licenseNote && <content />`

EthanFreestone (approved):

PR #831, opened 2021-07-12
ERM-1145 new agreement line UX improvements

CalamityC:
* add InfoPopover, Tooltip to BasketSelector
* tweak onAdd callback
* add setFieldData error for linkedResource field

### DIFF
--- a/src/components/AgreementLineSections/FormEresource.js
+++ b/src/components/AgreementLineSections/FormEresource.js
@@ -94,9 +94,9 @@ const FormEresource = ({
               label={<FormattedMessage id="ui-agreements.eresource" />}
               name={input.name}
               onAdd={resource => {
-                if (isEmpty(resource.id)) {
+                if (!resource.id) {
                   setFieldData('linkedResource', { error: <FormattedMessage id="ui-agreements.basketSelector.selectResourceMessage" /> });
-                  input.onChange();
+                  input.onChange(resource);
                   return;
                 }
                 input.onChange(resource);
--- a/src/components/AgreementLineSections/FormEresource.js
+++ b/src/components/AgreementLineSections/FormEresource.js
@@ -64,19 +64,17 @@ const FormEresource = ({
 
   // validation fires when there is no description and no eresource
   const required = (val, allValues) => {
-    console.log(val, 'val');
     if (allValues.description?.length > 0 || (!isEmpty(val) && !isDetached(val))) {
       return undefined;
     }
-    // return <FormattedMessage id="ui-agreements.agreementLine.provideEresource" />;
-    return 'validation message from FormEresource';
+    return <FormattedMessage id="ui-agreements.agreementLine.provideEresource" />;
   };
 
   return (
     <Field name="linkedResource" validate={required}>
       {({ input, meta }) => {
         const res = isExternal(input.value) ? input.value : (input.value.resource?._object ?? {});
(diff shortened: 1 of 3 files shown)


src/components/AgreementLineSections/FormEresource.js
```diff
@@ -90,10 +90,15 @@ const FormEresource = ({
               addButtonLabel={<FormattedMessage id="ui-agreements.agreementLine.linkSelectedEresource" />}
               basket={basket}
               component={BasketSelector}
-              error={meta.touched && meta.error}
+              error={(meta.touched && meta.error) || meta.data.error}
               label={<FormattedMessage id="ui-agreements.eresource" />}
               name={input.name}
               onAdd={resource => {
+                if (isEmpty(resource.id)) {
```
aditya-matukumalli, line 97: This could just be 
```suggestion
                if (!resource.id) {
```

src/components/AgreementLineSections/FormEresource.js
```diff
@@ -90,10 +90,15 @@ const FormEresource = ({
...
+                  setFieldData('linkedResource', { error: <FormattedMessage id="ui-agreements.basketSelector.selectResourceMessage" /> });
+                  input.onChange();
```
aditya-matukumalli, line 99: You could still pass in the resource to `onChange` right? Is that causing an issue?

src/components/BasketSelector/BasketSelector.js
```diff
@@ -67,13 +82,39 @@ export default class BasketSelector extends React.Component {
...
+            </div>
+            )}
+        </Tooltip>
+      );
+    }
     return (
       <Button
         buttonStyle="primary"
-        disabled={!item.id}
+        // disabled={!item.id}
```
aditya-matukumalli, line 117: commented out?

aditya-matukumalli (changes requested):

src/components/BasketSelector/BasketSelector.js
CalamityC: removed it

src/components/AgreementLineSections/FormEresource.js
CalamityC: yes, that's causing an error. I thought I fixed it with omitting the resource, but that's not the case. 
When I click the "Link selected e-resource" button without having selected an item, the "Save & close" button becomes active. If I then click "Save & close" I get an error.
Bildschirmfoto 2021-07-13 um 12 15 33

CalamityC: So this error ^^^ is still unresolved. Any ideas?

aditya-matukumalli: The `onAdd` callback should give you an empty object with `{id: ''}`. Calling `input.onChange(resource}` sets the value of the `Field` to be `{id: ''}`. You could then check if there is an `id` on the value within the `validate` function and if there is an id key and its value is an empty string you can return the custom error you want to.

aditya-matukumalli: The save and close isnt disable in this case here because we are setting the error at `meta.data.error` rather than `meta.error`. Save and close is disabled only if the `meta.error` value is set. So, current solution of `setFieldData` shouldn't be used in this case.

aditya-matukumalli (approved):

PR #866, opened 2021-09-07
Erm 1345 Added test coverage for AgreementLines Coverage component

MonireRasouli:


src/components/AgreementLineSections/Coverage/Coverage.test.js
```diff
@@ -0,0 +1,392 @@
...
+    'class': 'org.olf.kb.Pkg'
+    },
+    'addedTimestamp': 1630980265534,
+    'accessEnd': '2021-03-31',
+    'name': "'\"Institutions, industrial upgrading, and economic performance in Ja...' on Platform 'Elgaronline' in Package Edward Elgar:Edward Elgar E-Book Archive in Business & Management, ...",
+    'lastSeenTimestamp': 1630980265534,
+    'suppressFromDiscovery': false,
+    'longName': "'\"Institutions, industrial upgrading, and economic performance in Japan: the 'flying-geese' paradigm of catch-up growth\"' on Platform 'Elgaronline' in Package Edward Elgar:Edward Elgar E-Book Archive in Business & Management, Economics and Finance:Nationallizenz",
+    'class': 'org.olf.kb.PackageContentItem'
+};
```
aditya-matukumalli, line 370: Lets move this static data into a `testResources` file and import it frmo there.

src/components/AgreementLineSections/Coverage/Coverage.test.js
```diff
@@ -0,0 +1,392 @@
...
+
+describe('Coverage', () => {
+      beforeEach(() => {
+         renderWithIntl(
+           <MemoryRouter>
```
aditya-matukumalli, line 375: Do we need the `MemoryRouter` for these tests, i.e. I don't see a `Link` component being used anywhere in the component. Please double check and take it out if unnecessary.

src/components/AgreementLineSections/Coverage/Coverage.test.js
```diff
@@ -0,0 +1,392 @@
...
+      test('renders Coverage Accordion', async () => {
+        await Accordion('Coverage').exists();
+      });
+      test('renders expected column count', async () => {
+        await MultiColumnList({ columnCount: 7 }).exists();
+      });
+      test('renders expected columns', async () => {
+        await MultiColumnList({ columns: ['Start date', 'Start volume', 'Start issue', 'End date', 'End volume', 'End issue', 'Coverage type'] }).exists();
+      });
+});
```
aditya-matukumalli, line 391: This component needs more tests, i.e. checking individual cells of the MultiColumnList to see if the displayed values in the MCL columns are what we expect them to be. Refer to the [AllPeriods](https://github.com/folio-org/ui-agreements/blob/master/src/components/AgreementSections/AllPeriods/AllPeriods.test.js#L72-L73) tests to find the MultiColumnListCell interactor usage. You can use this interactor to target individual MCL cells content

aditya-matukumalli (changes requested):

aditya-matukumalli:
Please update the title of this PR to _Added test coverage for AgreementLines Coverage component_ so that its easy to find it when searching through the PRs.

src/components/AgreementLineSections/Coverage/Coverage.test.js
```diff
@@ -0,0 +1,81 @@
...
+
+  test('renders expected coverage end issue in the row', async () => {
+    Promise.all([
+      await MultiColumnListCell({ row: 0, columnIndex: 5 }).has({ content: '09/28/2021' }),
+    ]);
+  });
+
+  test('renders expected coverage type in the row', async () => {
+    Promise.all([
+      await MultiColumnListCell({ row: 0, columnIndex: 6 }).has({ content: 'Custom ' }),
```
aditya-matukumalli, line 77: We can take out the `Pomise.all` usage from these tests, as we are awaiting only a single promise in all of them. We generally make sure we have atleast 2 sets of data (so, 2 rows within the MCL) in the tests and in those scenarios you would have to use `Promise.all` if necessary. So, you could either extend these tests to test 2 resources or if you think one data set it good enough just take out the Promise.all usage.

aditya-matukumalli (changes requested):

src/components/AgreementLineSections/Coverage/Coverage.js
```diff
@@ -12,8 +12,6 @@ import {
 
 import { Embargo, isPackage } from '@folio/stripes-erm-components';
 
-import CustomCoverageIcon from '../CustomCoverageIcon';
```
aditya-matukumalli, line 15: Was removing the CustomCoverageIcon part of some JIRA?

aditya-matukumalli (changes requested):

src/components/AgreementLineSections/Coverage/Coverage.js
MonireRasouli: No, but it was causing the test to fail, we discussed that with Gill and Ethan, and they suggested that to remove it as it was not necessary to have the icon next to the custom coverage content.

aditya-matukumalli: Ah okay, sounds good.

src/components/AgreementLineSections/Coverage/Coverage.test.js
```diff
@@ -0,0 +1,157 @@
...
+          <Coverage line={line} resource={resource} />,
+          translationsProperties
+        );
+      });
+
+      test('renders Coverage Accordion', async () => {
+        await Accordion('Coverage').exists();
+      });
+
+      test('renders Coverage Accordion', async () => {
```
aditya-matukumalli, line 22: same test as above

src/components/AgreementLineSections/Coverage/Coverage.test.js
```diff
@@ -0,0 +1,157 @@
...
+        Promise.all([
+          await MultiColumnListCell({ row: 0, columnIndex: 4 }).has({ content: '8' }),
+          await MultiColumnListCell({ row: 1, columnIndex: 4 }).has({ content: '8' })
+        ]);
+      });
+
+      test('renders expected coverage end issue in the row', async () => {
+        Promise.all([
+          await MultiColumnListCell({ row: 0, columnIndex: 5 }).has({ content: '12' }),
+          await MultiColumnListCell({ row: 1, columnIndex: 5 }).has({ content: '12' })
```
aditya-matukumalli, line 86: Can we have different content values for individual rows, because having the same content value for both the rows could cause false positives

aditya-matukumalli (changes requested):

src/components/AgreementLineSections/Coverage/Coverage.test.js
MonireRasouli: sure

MonireRasouli: I changed the values.

src/components/AgreementLineSections/Coverage/Coverage.test.js
```diff
@@ -0,0 +1,157 @@
...
+          <Coverage line={line} resource={resource} />,
+          translationsProperties
+        );
+      });
+
+      test('renders Coverage Accordion', async () => {
+        await Accordion('Coverage').exists();
+      });
+
+      test('renders Embargo Accordion', async () => {
```
aditya-matukumalli, line 22: Description says Accordion but the test is checking for a KeyValue?

aditya-matukumalli (changes requested):

src/components/AgreementLineSections/Coverage/Coverage.test.js
MonireRasouli: I fixed that

PR #868, opened 2021-09-08
Erm 1346 Added tests for AgreementLineSection EresourceSelector component

MonireRasouli:


src/components/AgreementLineSections/EresourceSelector/EresourceSelector.test.js
```diff
@@ -0,0 +1,40 @@
...
+    const { getByText } = renderComponent;
+    expect(getByText('eHoldings')).toBeInTheDocument();
+   });
+
+   test('renders linkEresourceToStart message', () => {
+    const { getByText } = renderComponent;
+    expect(getByText('Link an e-resource to get started')).toBeInTheDocument();
+   });
+
+   test('renders error message', () => {
```
aditya-matukumalli, line 35: This test needs to be broken into a different describe block. This block should deal with rendering an EresourceSelector component with a line. The second describe block should also pass in an error. Lets not mix up both the tests into one block

aditya-matukumalli (changes requested):

aditya-matukumalli (approved):

PR #879, opened 2021-09-20
ERM-1280 Add test coverage for ui-agreements LinkedLicenseCard

CalamityC:
add tests for LinkedLicenseCard component, ERM-1280


src/components/LinkedLicenseCard/LinkedLicenseCard.test.js
```diff
@@ -0,0 +1,101 @@
...
+        const { getByRole } = renderComponent;
+        expect(getByRole('link', { name: futureLicense.remoteId_object.name })).toBeInTheDocument();
+      });
+
+      test('renders the LicenseCard', () => {
+        const { getByText } = renderComponent;
+        expect(getByText('LicenseCard')).toBeInTheDocument();
+      });
+
+      if (futureLicense.note) {
```
aditya-matukumalli, line 46: We dont need to add the conditional to run a test here. Component needs to be tested based on the license prop passed to the `LinkedLicenseCard` component.

So, in this case, as we are defining the test resource, we already know if this test should pass or fail. Lets just test for that. You shouldnt need the if condition

src/components/LinkedLicenseCard/LinkedLicenseCard.test.js
```diff
@@ -0,0 +1,101 @@
...
+        test('renders the expected note value', async () => {
+          await KeyValue('Note').has({ value: futureLicense.note });
+        });
+      }
+
+      if (futureLicense.remoteId_object.amendments.length) {
```
aditya-matukumalli, line 52: same comment as above

src/components/LinkedLicenseCard/LinkedLicenseCard.test.js
```diff
@@ -0,0 +1,101 @@
...
+      });
+
+      test('renders LinkedLicenseCard component', () => {
+        const { getByTestId } = renderComponent;
+        expect(getByTestId('linkedLicense')).toBeInTheDocument();
+       });
+
+       test('renders a link with the license name', () => {
+        const { getByRole } = renderComponent;
+        expect(getByRole('link', { name: futureLicense.remoteId_object.name })).toBeInTheDocument();
```
aditya-matukumalli, line 38: Lets just test the actual expected value instead of the accessing it vai the properties of the `futureLicense`. That way it will be easy when reading through these tests. Please change them across this file

aditya-matukumalli (changes requested):

aditya-matukumalli (approved):

PR #889, opened 2021-09-29
ERM-1566 Agreement lines record count badge disappears when the accor…

CalamityC:
…dion is opened

* add renderBadge function to renderAddAgreementLineButton, ref ERM-1566


src/components/AgreementSections/Lines/Lines.js
```diff
@@ -31,11 +31,14 @@ export default class Lines extends React.Component {
 
   renderAddAgreementLineButton = () => {
```
aditya-matukumalli, line 32: Can we change the name of the function to maybe `renderAddAgreementLineButtonAndBadge` as we have now extended this component

aditya-matukumalli (changes requested):

aditya-matukumalli (approved):

PR #896, opened 2021-10-05
Erm 1401

MonireRasouli:


src/components/PlatformSections/PlatformInfo/PlatformInfo.test.js
```diff
@@ -0,0 +1,25 @@
...
+      translationsProperties
+    );
+  });
+
+  test('renders the expected locators key value', async () => {
+    await KeyValue('Locators').has({ value: 'msp.org' });
+  });
+
+  test('renders the expected local platform code key value', async () => {
+    await KeyValue('Local platform code').has({ value: '-' });
```
aditya-matukumalli, line 22: Can we actually test for a valid platform code instead of testing for a noValue?

aditya-matukumalli (changes requested):

PR #897, opened 2021-10-06
fix: Empty index.js

EthanFreestone:
Filled an empty index.js file for UsageData that was causing crashes on agreements with usageDataProviders

ERM-1887


MikeTaylor (approved):
Yes, please!

PR #914, opened 2021-10-25
Erm 1472, Added tests for AgreementViewRoute

MonireRasouli:
*Mock buttons to handle callbacks in Route
*Mock fetches


src/routes/AgreementViewRoute/AgreementViewRoute.js
```diff
@@ -206,6 +206,7 @@ class AgreementViewRoute extends React.Component {
 
   static contextType = CalloutContext;
 
+  /* istanbul ignore next */
   componentDidUpdate(prevProps) {
```
aditya-matukumalli, line 210: You can use the `rerender` helper from RTL to trigger the componentDidUpdate. Lets avoid using `istanbul ignore next` here

example usage: https://github.com/folio-org/ui-agreements/blob/master/src/routes/AgreementsRoute/AgreementsRoute.test.js#L226

aditya-matukumalli (changes requested):
Please remove the `istanbul ignore next` comments and lets try to cover these functions within the tests

src/routes/AgreementViewRoute/AgreementViewRoute.js
```diff
@@ -214,6 +214,7 @@ class AgreementViewRoute extends React.Component {
     }
   }
 
+  /* istanbul ignore next */
```
aditya-matukumalli, line 217: Any reason you are ignoring all the callbacks here for code coverage? All of these callbacks can be covered via the tests. Please double check.

aditya-matukumalli (changes requested):

aditya-matukumalli (approved):

PR #944, opened 2021-11-18
ERM-1400 Add tests for POLinesFieldArray component

MonireRasouli:


aditya-matukumalli (changes requested):
We need more tests here. Please refer to other FieldArray tests we have (good example is the [LicensesFieldArray](https://github.com/folio-org/ui-agreements/blob/master/src/components/LicensesFieldArray/tests/LicensesFieldArray.test.js) tests. Needs additional tests for initialValues too.

aditya-matukumalli (approved):

PR #974, opened 2021-12-13
ERM-1953 Re-implement display of acquistionsMethods following order-l…

MonireRasouli:
…ines 3.0 interface


src/components/POLineCard/POLineCard.js
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
+// eslint-disable-next-line import/no-extraneous-dependencies
+import { useAcqMethods } from '@folio/orders/src/common/hooks/useAcqMethods/useAcqMethods';
```
EthanFreestone, line 5: If this import can be replaced by [this](https://github.com/folio-org/stripes-acq-components/blob/master/lib/hooks/useAcqMethodsOptions/useAcqMethodsOptions.js) then that should happen. We don't want to rely on @folio/orders

src/components/utilities/getTranslatedAcqMethod.js
```diff
@@ -0,0 +1,16 @@
+/* eslint-disable import/prefer-default-export */
+import { invert } from 'lodash';
```
EthanFreestone, line 2: Please change this to ```import invert from 'lodash/invert'``` and retest

EthanFreestone (changes requested):

EthanFreestone (approved):

PR #983, opened 2021-12-17
Update "source-manager-job-executions" interface version

RuslanLavrov:
## Purpose
In the scope of [PR](https://github.com/folio-org/mod-source-record-manager/pull/530/commits/8089c6aa8e42ca7d63d3c8990128eb1cffabdc18) the query parameters for `GET /metadata-provider/jobExecutions` endpoint were changed, so 
"source-manager-job-executions" interface version will be increased from 2.0 to 3.0. It is necessary to update interface dependency

## Approach
* update "source-manager-job-executions" version from 2.0 to 3.0

## Learning


KaterynaSenchenko:
Hi @EthanFreestone, could you please merge this PR if you think the changes are necessary - we bumped the "source-manager-job-executions" version from 2.0 to 3.0 and going through all the modules that reference this API

PR #988, opened 2022-01-12
wrap activeFrom and activeTo dates in FormattedUTCDate component, ref…

CalamityC:
…s ERM-1976


EthanFreestone (changes requested):
The tests were wrongly created expecting the broken display. Please amend those so they expect the correct date

EthanFreestone (approved):

PR #1000, opened 2022-02-25
fix: Order line fetch

EthanFreestone:
Fixed order line fetching so that when a new POLine is created or removed, the subsequent render of the agreement itself reflects that


EthanFreestone:
refs ERM-1999

PR #1052, opened 2022-06-21
ERM-2215: Refactor EditableRefdataList/EditableRefdataCategoryList

EthanFreestone:


EthanFreestone:
This is just about ready to go, but will rely on kint-components 2.7.0

PR #1110, opened 2022-09-21
ERM-2214, Migrate ERM Picklist Settings to new EditableRefdataList co…

MonireRasouli:
…mponent


src/settings/routes/PickListValueSettings.js
```diff
@@ -37,6 +37,9 @@ const PickListValues = () => {
           label={
             <FormattedMessage id="ui-agreements.settings.pickListValues" />
           }
+          labelOverrides={{
```
EthanFreestone, line 40: This will not run as things stand, it _should_ have crashed your frontend when you tried to run it... did it not?

You need to make sure that you have the latest `ui-agreements` code and that you've nuked your workspace.

If you're currently building with kint-components in your workspace that also needs to be up to date.


Either way, this function can be replaced by:
```deleteError: 'ui-agreements.settings.pickListValues.deletePickListValueError'``` and you'll need to change the translation to use `label` and `error` instead of `name` and `error`. 

Make sure whenever you make changes that you're testing them locally, not just checking if tests pass, because this code as it stands should crash

EthanFreestone (changes requested):

EthanFreestone (approved):

PR #1150, opened 2022-11-15
ERM-2445, Display a callout when exporting an Agreement

MonireRasouli:


src/routes/AgreementViewRoute/AgreementViewRoute.js
```diff
@@ -273,6 +273,10 @@ const AgreementViewRoute = ({
     });
   };
 
+  const exportAgreementAsJSON = () => {
+    callout.sendCallout({ type: 'success', message: <FormattedMessage id="ui-agreements.agreements.exportingAgreement" /> });
+  };
```
EthanFreestone, line 278: I don't like the naming here.

Sending the callout should happen in a `.then` on the onExportAgreement. We shouldn't need an extra method to call the callout, especially because this callout will fire whether the query was successful or not.

The naming also implies that we're making a second export, which is not what this method is doing.

EthanFreestone (changes requested):

EthanFreestone (approved):

PR #1158, opened 2022-11-21
ERM-2451 / ERM-2416 Separate title search and package search in local KB e-resource searching

CalamityC:
* add Titles route and view component
* add route to index, add titles url
* fix onClose handler and row click location for titles
* fix titles MCL


CalamityC:
merged two branches and raised  #1161

PR #1189, opened 2023-02-03
ERM-2612, Add support for content type property on Agreements

MonireRasouli:


EthanFreestone:
We're using a different UX pattern now

PR #1196, opened 2023-03-01
ERM-2612: Add support for content type property on Agreements

MonireRasouli:


src/components/AgreementSections/ContentTypesFieldArray/ContentTypesFieldArray.js
```diff
@@ -0,0 +1,146 @@
...
+      contentTypeRefs.current.push(node);
+    }
+  };
+
+  const renderContentTypes = () => (
+    items.map((act, index) => {
+      const dataOptions = agreementContentType.filter(ct => !contentTypeInUse.includes(ct.value) || ct.value === act.contentType?.value);
+      return (
+        <div
+          key={act + index}
```
CalamityC, line 87: I'd prefer template literals here (like eg. in line 98) 
```suggestion
          key={`${act}${index}`}
```

EthanFreestone (changes requested):
Looks good, can we try to make sure all test refdata though goes to the jest helper file and is then parsed out from there for tests.


We're trying to push towards consistent testing objects, and refdata is the only one we're actually doing right now, so I don't want that to slip

EthanFreestone:
Ah, in addition, that code smell is pointing at a place where you create a key via concatentation. Please instead create a string like
```
 key={${act?.value}[${index}]}
``` or something. (I don't know the shape of act, if it's already a string then that's fine, but if it's an object it shouldn't be going into the key as is.

File: src/components/AgreementSections/ContentTypesFieldArray/ContentTypesFieldArray.js (hunk 1 above)
Comment (CalamityC) [reply src/components/AgreementSections/ContentTypesFieldArray/ContentTypesFieldArray.js:87]: rather use Ethans suggestion as`act` is an object, isn't it?

File: src/components/AgreementSections/ContentTypesFieldArray/ContentTypesFieldArray.js [hunk 2]
```diff
@@ -0,0 +1,146 @@
...
+      contentTypeRefs.current.push(node);
+    }
+  };
+
+  const renderContentTypes = () => (
+    items.map((act, index) => {
+      const dataOptions = agreementContentType.filter(ct => !contentTypeInUse.includes(ct.value) || ct.value === act.contentType?.value);
+      return (
+        <div
+          key={`${act?.contentType?.id}`}
```
CalamityC, line 87: That's fine I think. Although you're using here the refdata id (the contentType object is the refdata object), you could also use `act.id` which is the specific id of the agreementContentType object. But as we're not allowing to set a refdata value twice on an agreement, your `key` is also unique per agreement, so all should be good.

CalamityC (approved):

EthanFreestone (approved):

PR #1216, opened 2023-05-17
ERM-2421: Remove agreement line in-line edit

Jack-Golding:
Removed agreement line in-line edit from agreement edit view and associated tests

[ERM-2421](https://issues.folio.org/browse/ERM-2421)


EthanFreestone (changes requested):
Can we move the agreement post/put to the modal instead of the basket route?

PR #1255, opened 2023-10-02
ERM-2983: Documents filter

Jack-Golding:
Implemented new filter builder for documents relating to agreements

[ERM-2983](https://issues.folio.org/browse/ERM-2983)


src/components/utilities/selectifyRefdata.js
```diff
@@ -0,0 +1,25 @@
+const selectifyRefdata = (
```
EthanFreestone, line 1: Please also remove this, I have merged the erm-components PR adding this there

EthanFreestone (changes requested):
Extra change, for removing selectify refdata in favour of centralised version in stripes-erm-components

PR #1259, opened 2023-10-06
ERM-2054, Platforms Tab - Implement MCL Next/Previous pagination

MonireRasouli:
Implemented Next/Prev pagination
[ERM-2054](https://issues.folio.org/browse/ERM-2054)


EthanFreestone (approved):
Leaving comment here @MonireRasouli , have only just noticed the page PropType. That is not necxessary since we aren't passing a page prop. Merging this for expediency but can you bear in mind for future work, and perhaps remove those in a future PR? Apologies for only just catching that now

PR #1303, opened 2024-03-25
ERM-3172 The failure of reference object when displaying agreement li…

CalamityC:
…ne should be handled elegantly

* use ErrorCard component in AgreementLineSections Info
* print resource.reference in LinesList conditionally


CalamityC:
I generalized ErrorCard. Now I think it should reside in `stripes-erm-components`
What do you think @EthanFreestone ?

CalamityC:
@EthanFreestone
 I moved `ErrorCard` , so please merge https://github.com/folio-org/stripes-erm-components/pull/673 first
 I'll push changes and un-draft this PR as soon as that happened.

EthanFreestone (approved):

PR #1307, opened 2024-04-10
ERM-3186 Change default search options for Local KB titles to exclude…

CalamityC:
… identifiers

* remove identifiers.identifier.value from defaultTitlesQIndex
* add identifiers.identifier.value to the default searchKey (no checkboxes checked, no qIndex)
* set up the 'default' for qIndexSASQProps in Titles view

### DIFF
--- a/src/routes/TitlesRoute/TitlesRoute.js
+++ b/src/routes/TitlesRoute/TitlesRoute.js
@@ -65,7 +65,13 @@ const TitlesRoute = ({
   const { query, querySetter, queryGetter } = useKiwtSASQuery();
 
   const { currentPage } = usePrevNextPagination();
-  const { searchKey } = useSASQQIndex({ defaultQIndex });
+
+  /*
+   * Have the default searchKey (When no qIndex is present, meaning no checkboxes are checked) include identifier values
+   * This usage of useSASQQIndex is _not_ setting the qindex in the url,
+   * that is handled in the view via SASQ and the props handed to it.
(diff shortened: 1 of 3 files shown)


src/routes/TitlesRoute/TitlesRoute.js
```diff
@@ -65,7 +65,13 @@ const TitlesRoute = ({
...
 
   const { currentPage } = usePrevNextPagination();
-  const { searchKey } = useSASQQIndex({ defaultQIndex });
+
+  /*
+   * Have the default searchKey (When no qIndex is present, meaning no checkboxes are checked) include identifier values
+   * This usage of useSASQQIndex is _not_ setting the qindex in the url,
+   * that is handled in the view via SASQ and the props handed to it.
+   */
+  const { searchKey } = useSASQQIndex({ defaultQIndex: `${defaultQIndex},identifiers.identifier.value` });
```
EthanFreestone, line 74: As of this PR: https://github.com/folio-org/stripes-erm-components/pull/675

We can now make this clearer in the split-hook use case by changing `defaultQIndex` for `defaultSearchKey`.

This is not _necessary_ since in the absence of that prop it falls back to defaultQIndex, so I'm approving this PR anyway, but we could choose to clean these up in future if we wished and thought it provided more maintainable/understandable code. Out of scope though for this PR/Issue

EthanFreestone (approved):

PR #1331, opened 2024-06-26
ERM-3183 Create agreement line form does not clear on "Save and creat…

CalamityC:
…e another"

* add minimal delay between pushing the view page and opening the create page again


EthanFreestone (approved):
I think I approve this as a fix, but I reckon we should probably look into another way to ensure the form reinitialises...

Depends on if this needs backporting or not as to whether we merge here or close and reexamine

EthanFreestone (changes requested):
Can we change this to use the form.reset mechanism we do elsewhere please?

From Create we need to reset the form, from edit we should redirect to "create" (which should clear the form for us)

EthanFreestone (approved):

PR #1344, opened 2024-08-05
ERM-3277 Add Material Type filter (Print/Electronic) to Title search

CalamityC:
* add subType filter
* name it material type

### DIFF
--- a/src/components/TitleFilters/TitleFilters.js
+++ b/src/components/TitleFilters/TitleFilters.js
@@ -12,7 +12,7 @@ const propTypes = {
 };
 
 const FILTERS = [
-  'materialType',
+  'subType',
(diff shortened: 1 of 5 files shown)


EthanFreestone (changes requested):
Thanks @CalamityC, changes look great, but I'd rather the code reflect the names as per in the backend, so could materialType be changed for `subType`, with just the translation differing?

This is a matter of taste, it makes grepping slightly more difficult if a frontend facing user reports an issue, but makes it clearer I think to newer developers where the values come from etc. I'd rather keep the amount of alternative names developers have to understand "materialType == subType" etc to a minimum :)

EthanFreestone (approved):

PR #1352, opened 2024-10-04
ERM-3362: Error on viewing ERM basket

EthanFreestone:
fix: Fixes to accommodate stripes changes

Stripes changed to use `formatter` prop in Selection for rendering display as well as the options. This means that a `null` option is now a legitimate use case, which we had not protected for. Added in null safety here, and this also means that the display on selection is equivalent to the options render, which might not be desired but at least this works.

ERM-3362

### DIFF
--- a/src/components/views/Basket/Basket.js
+++ b/src/components/views/Basket/Basket.js
@@ -200,25 +200,30 @@ const Basket = ({
         <Col md={8} xs={12}>
           <Selection
             dataOptions={openAgreementsState}
-            formatter={({ option }) => (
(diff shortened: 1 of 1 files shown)


EthanFreestone:
Stripes change found here:
https://github.com/folio-org/stripes-components/commit/4349f109bb5b2c4d5de2bd59a66d3a78fa2b0e34#diff-1673a9437bbad5531a73a22cb154830fbf4ed6beeca982b29ef58d2fe8be2f24R215

PR #1361, opened 2024-10-21
ERM-3332

MonireRasouli:


MonireRasouli:
should be merged after stripes-erm-component [ERM-3332](https://github.com/folio-org/stripes-erm-components/pull/704)

PR #1366, opened 2024-11-04
ERM-3390 Show columns: Name column should not be in the list of unsel…

CalamityC:
…ectable columns

* exclude 'name' column from toggle menu


EthanFreestone (approved):

EthanFreestone (changes requested):
Tests failing because ColumnManager is now mocked by default. Easiest fix would be to use `jest.requireActual` for this test, but I leave that up to you @CalamityC

EthanFreestone (approved):

PR #1368, opened 2024-11-11
ERM-3392 Changes to the field content type are not saved, but a success message is displayed

CalamityC:
* change agreementContentTypes initialValues on submit


EthanFreestone (changes requested):
This isn't right. This will disregard any _delete instructions.


We should be able to instead have the actual _form_ manipulate the id instead, so there's nofanagling needed on submit.

PR #1383, opened 2025-01-10
Erm 3446, Agreements: Basket should use the agreements look-up component instead of single select search

MonireRasouli:
[ERM-3446](https://folio-org.atlassian.net/jira/software/c/projects/ERM/boards/141?selectedIssue=ERM-3446)


src/components/views/Basket/Basket.js
```diff
@@ -161,96 +164,21 @@ const Basket = ({
...
-          </Button>
-        </Layout>
-      </div>
-    );
-  };
-
-  const renderAddToAgreementSection = () => {
-    if (!openAgreementsState.length) return null;
+  const renderAddToSelectedAgreementButton = () => {
+    const disabled = Object.values(selectedItems).find((v) => v) === undefined;
```
EthanFreestone, line 168: Again, this takes out consideration for the `selectedAgreementId` stat, but we haven't actually gotten rid of that logic.

Either we're changing the pattern to no longer need selectedAgreementId (adn it should be removed entirely) or we're using that pattern (and so shouldn't be removing parts of functionality)

src/components/views/Basket/Basket.js
```diff
@@ -280,7 +208,7 @@ const Basket = ({
             selectedItems={selectedItems}
           />
           {renderCreateAgreementButton()}
-          {renderAddToAgreementSection()}
+          {renderAddToSelectedAgreementButton()}
```
EthanFreestone, line 211: See comment on line 167

src/components/views/Basket/Basket.test.js
```diff
@@ -81,62 +79,6 @@ describe('Basket', () => {
     await Button('Create new agreement').exists();
   });
 
-  describe('Choosing an option and clicking the add to selected agreement button', () => {
```
EthanFreestone, line 84: Removing a load of tests and not replacing them is not really acceptable. We should be replacing the functionality with a mock where possible to ensure that our module does the right thing when that mocked component (not agreementFilterButton) has inputs/outputs

EthanFreestone (changes requested):
This doesn't look right to me I'm afraid, sorry :(


We have a mix of two different patterns in play, the old pattern (set agreement Id in state and then make PUT to that) and a new pattern, which seems to set up a different PUT method and then not use it? In addition it seems to me like selecting an agreement will currently _immediately_ fire the PUT and add the items to that agreement, which is definitely not what we want.

This overall looks confusing and over complicated, I think it's a case of copying code/patterns from elsewhere without first understanding what they are doing.

In general I would expect a change of logic to be deliberate, and explained in comments/the PR.

src/components/views/Basket/Basket.js
```diff
@@ -161,96 +164,21 @@ const Basket = ({
...
-        </Col>
-        {renderAddToAgreementButton()}
-      </div>
+      <Layout className="marginTop1">
+        <AgreementFilterButton
+          buttonStyle="primary"
+          disabled={disabled}
+          name="agreement"
+          onAgreementSelected={(agreement) => {
+            setSelectedAgreementId(agreement.id);
```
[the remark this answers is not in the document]
MonireRasouli: again I'm not sure which save button you're referring to. I believe we only have a button for selecting an agreement, which opens the existing agreements. Then when an agreement is selected, it should link the resource to the agreement and open the agreement with its information. There is no save button involved in this process.

src/components/views/Basket/Basket.js
```diff
@@ -161,96 +164,21 @@ const Basket = ({
...
+            handleAddToExistingAgreement(agreement.id, getSelectedItems());
```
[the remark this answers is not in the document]
MonireRasouli: I'm not sure which save button you are referring to. Could you please clarify which save button you mean? I want to make sure I understand your concern correctly.

EthanFreestone (approved):

EthanFreestone (approved):

EthanFreestone (approved):

PR #1385, opened 2025-02-10
ERM-3568, Add package synchronise property in UI

MonireRasouli:


src/components/EResourceSections/PackageInfo/PackageInfo.js
```diff
@@ -41,6 +41,17 @@ const PackageInfo = ({
     _object: eresource,
   };
 
+  let synchronisationStatus;
+  if (eresource?.syncContentsFromSource !== null) {
```
EthanFreestone, line 45: This is a lot of additional logic, and it's repeated below.

<NoValue> is already rendered by `keyvalue`...
https://github.com/semantic-release/commit-analyzer/blob/master/lib/default-release-rules.js#L30



So you can probably instead do something like:
```
<KeyValue
  label={<FormattedMessage id="ui-agreements.eresources.SynchronisationStatus" />}
  value={eresource?.syncContentsFromSource !== null ? intl.formatMessage({ id: `ui-agreements.eresources.syncContentsFromSource.${eresource.syncContentsFromSource}` }) : ''}
 />
```
and then make sure your translations are:
```
"eresources.syncContentsFromSource.true": "Synchronising",
"eresources.syncContentsFromSource.false": "Paused"
```

Given that this is repeated in two places, I'd also probably separate this out to its own utility function (passing in intl from useIntl) or (better) a component returning the keyvalue (and using useIntl itself).

src/components/EResourceSections/PackageInfo/PackageInfo.js
```diff
@@ -148,7 +159,17 @@ const PackageInfo = ({
...
       </Row>
-      <PackageIdentifiers pkg={eresource} />
+      <Row>
+        <Col xs={6}>
+          <PackageIdentifiers pkg={eresource} />
+        </Col>
+        <Col xs={3}>
+          <KeyValue
+            label={<FormattedMessage id="ui-agreements.eresources.reference" />}
+            value={eresource.reference || <NoValue />}
```
EthanFreestone, line 169: See above, this is unnecessary since NoValue is already built into KeyValue

src/components/PackageCard/PackageCard.js
```diff
@@ -39,6 +39,17 @@ const PackageCard = ({
...
 
+  let synchronisationStatus;
+  if (pkgObject?.syncContentsFromSource !== null) {
+    if (pkgObject?.syncContentsFromSource) {
+      synchronisationStatus = <FormattedMessage id="ui-agreements.eresources.synchronising" />;
+    } else {
+      synchronisationStatus = <FormattedMessage id="ui-agreements.eresources.paused" />;
+    }
+  } else {
+    synchronisationStatus = <NoValue />;
```
EthanFreestone, line 50: See above... I don't want us to repeat this kind of logic, when we make changes these things break. Utility component please `PackageSyncronisingValue` or something

src/components/PackageCard/PackageCard.js
```diff
@@ -97,7 +108,16 @@ const PackageCard = ({
...
         </Col>
       </Row>
-      <PackageIdentifiers pkg={pkgObject} />
+      <Row>
+        <Col xs={6}>
+          <PackageIdentifiers pkg={pkgObject} />
+        </Col>
+        <Col xs={3}>
+          <KeyValue label={<FormattedMessage id="ui-agreements.eresources.reference" />}>
+            <div data-test-package-reference>{pkgObject?.reference ?? <NoValue />}</div>
```
EthanFreestone, line 117: NoValue isn't rquired inside KeyValue

EthanFreestone (changes requested):

PR #1388, opened 2025-02-17
ERM-3581 Support changing the sync status of a single package in the …

CalamityC:
…local KB

* convert Package component to functional

### DIFF
--- a/src/components/views/Package/Package.js
+++ b/src/components/views/Package/Package.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useRef } from 'react';
 import PropTypes from 'prop-types';
 import { FormattedMessage } from 'react-intl';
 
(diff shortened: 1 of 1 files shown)


EthanFreestone (approved):
I@m happy with this approach and willmerge, with 2 comments:

- The query key I'd like to use a slightly different pattern in general. It matters less for mutations anyway, so I'm happy to leave it, but I'd like us to namespace those with ['ERM', 'ResourceName', 'id', 'action', 'queryParams', 'anyghing else']

Keeping a consisteent pattern will help us when we wish to manipulate the query state later. Again though this is much more of an issue for GETs so I'm happy to merge for now, I believe we use the pattern here elsewhere as well.


Secondly I don't really understand why we have two functions for synchronise? At the moment we set up a function `synchronisePackage` and then call it in a _new_ function `handleSynchronize`. This isn't wrong per se, it's just a little confusing spreading the logic in two places.

EthanFreestone:
Whoops, nearly merged missing slack thread. My comments above stand, leaving for Monday to review with team :)

CalamityC:
> Secondly I don't really understand why we have two functions for synchronise? At the moment we set up a function `synchronisePackage` and then call it in a _new_ function `handleSynchronize`. This isn't wrong per se, it's just a little confusing spreading the logic in two places.

I took the example from agreement delete
I've no idea how to do it otherwise. If you have an advise, you're welcome.
Just moving the then / catch parts for the messages into the useMutation function didn't work for me.

EthanFreestone (approved):

PR #1389, opened 2025-02-18
ERM-3580, feat: Add checkbox functionality to package list

MonireRasouli:
- Implemented a checkbox in the header of the MultiColumnList to allow users to select/deselect all packages on the current page.
- Added individual checkboxes for each package row to enable selection of specific packages.
- Added Atction button for synchronising and pause the synchronising packages


EthanFreestone:
Approved, small changes made to:
- Prevent sorting by "select", since that's not a persisted value
- Tweak sync send and callout logic, since they were really doing the same thing, there was no need to separate those
- Change over to using only selectedPackageIds rather than 2 states

@MonireRasouli Good work on this, have a check hen you get a chance to double check these changes make sense to you and you don't have any questions/it won't affect your testing

PR #1392, opened 2025-02-21
refactor: move settings over to newer useSettings shape for neatness

EthanFreestone:


EthanFreestone:
In draft until stripes kint components is released as expected

PR #1394, opened 2025-03-03
ERM-3628 When there is a single search result in the agreements searc…

CalamityC:
…h, the results continually refresh

* change useEffect in AgreementsRoute


src/routes/AgreementsRoute/AgreementsRoute.js
```diff
@@ -127,9 +127,13 @@ const AgreementsRoute = ({ children, history, location, match }) => {
 
   useEffect(() => {
     if (agreementsCount === 1) {
-      history.push(`${urls.agreementView(agreements[0].id)}${location.search}`);
+      const newUrl = `${urls.agreementView(agreements[0].id)}${location.search}`;
```
EthanFreestone, line 130: I like the logic to check we're not overwriting with the same url repeatedly, but I would lower the threshold for "the same" to not include `${location.search}`.


ie newUrl is the url built from id, and the comparator is location.pathname, if they're not the same THEN history.push to `url(agreementId)${location.search}`

EthanFreestone (changes requested):

EthanFreestone:
This was marked as [cannot reproduce](https://folio-org.atlassian.net/browse/ERM-3628), so I'm going to close this PR for now

PR #1400, opened 2025-03-17
ERM-3652, Extend tests for Package multi-select/synchronise switch

MonireRasouli:


EthanFreestone (changes requested):
Hi @MonireRasouli I'm mostly happy with the workflow, and that we're testing the necessary things, although I think we ought to check that

- selecting single package ONLY selects that package, the other checkbox remains unclicked
- selecting all selects all rows (all checkboxes are clicked)


However we need to make sure that we're testing in a way which _ensures_ this works in all environments, which means we have a pattern to uphold.


Can you please change these tests to obey the 

- describe (some SINGLE action happens)
  - test (some SINGLE outcome has occurred)
  - more tests, if needed... one per outcome

Then if a new describe relates to a previous describe it happens inside the describe block, so the test cascades down.

EthanFreestone:
@MonireRasouli Thanks for the changes. This still doesn't quite fit the pattern I described though. The idea is that any kind of user interaction, say a button click, happens in a describe. 

We're not testing whether a button can be clicked, that's a Stripes issue, and behind that a react/html issue.

What we want to know in a unit test is whether the outcome of a button press is as expected.

This is the same for "fill in a form field", that is a user interaction, so belongs in a describe.

This way we can start to build up a nested test file, describing all unit interactions and testing all outcomes, which will boost the coverage as well as making it crystal clear what the expected outcome is. 

It _also_ makes it really easy then to swap between different test cases and test the same outcomes, so upping the test coverage even higher and chekcing that everything behaves the same in every diffferent possible permutation of starting conditions.


I know you're working on ILL now, and I'm very aware you're behind there thanks to me messing up the ILL environment, so I'm going to pick this up with expected changes and I'll ping you when they're in.

I'm very happy to have another jest workshop at some point, and on unit vs integration testing in general if we think it'd be helpful

PR #1402, opened 2025-03-31
ERM-3670 When package 'syncContentsFromSource' is not set, display with a '-'

CalamityC:
* fix if condition
* extend test


EthanFreestone (changes requested):
I'd prefer we used proper evaluation here rather than trying to guess type with an unchecked `!=`. It's smart but it introduces "magic" knowledge into the system.

TLDR, can we please do `if(syncContentsFromSource !== undefined && syncContentsFromSource !== null)`, I know it's more wordy but it makes things clearer when we come to maintain it later.

EthanFreestone (approved):

PR #1412, opened 2025-04-29
ERM-3712: Remove eresources.syncMultiplePackages* translation keys

julianladisch:
https://folio-org.atlassian.net/browse/ERM-3712

It seems the these translation keys are not used and should be deleted to reduce the translator's workload.


ostephens (changes requested):
These cannot be removed as they are used from the packages route https://github.com/folio-org/ui-agreements/blob/2470ab1c25711019c4d7fbeb3edbb466fe4d86b3/src/routes/PackagesRoute/PackagesRoute.js#L169

PR #1413, opened 2025-04-29
ERM-3711 Migrate all setting uses in ui-agreements to use app settings

CalamityC:
* switch to new endpoint in GeneralSettings
* add parse function
* change useAgreementsSettings


EthanFreestone:
Merged the stripes-erm-testing PR @CalamityC , sorry for the delay

EthanFreestone (approved):

PR #1415, opened 2025-05-13
ERM-3704, Hide tags icon when tags are disabled

MonireRasouli:


EthanFreestone:
Merged, see [linked comment](https://github.com/folio-org/stripes-erm-components/pull/728#issuecomment-2880447174) on stripes-erm-components PR

PR #1424, opened 2025-07-02
ERM-3743 Add support to delete the content of a single package to the UI

CalamityC:
* add permission
* add Button, Modals
* add translations
* add endpoints
* add handlers and toast message


EthanFreestone (approved):

EthanFreestone:
Approved and bumped up to master :)

The only things to note from me

-  the use of `<p>`: In general we tend to use <Layout> to get things where we want them to be, but it's really not a big deal so I'd be happy to merge.
- Inline coding: The various modals have been set up differently, one has it's content inline and one has it declared up above. Again it's not a blocker for merge, it's easy enough to read, but this _could_ be two confirmation modals only with different labels I think, and we could declare their messages etc etc in the same way.
  - In theory we could even move the specific rendering (and maybe state management?) for the two modals to their own component to make the code a little smoother to read, but again it's really not a big deal--left up to developer as to what is best

EthanFreestone:
> > * the use of `<p>`: In general we tend to use  to get things where we want them to be, but it's really not a big deal so I'd be happy to merge.
> 
> I do not understand what you mean with that.
> 
> > * Inline coding: The various modals have been set up differently, one has it's content inline and one has it declared up above. Again it's not a blocker for merge, it's easy enough to read, but this _could_ be two confirmation modals only with different labels I think, and we could declare their messages etc etc in the same way.
> 
> If you can give me a hint how to get the ConfirmationModal only render one Button, I'm happy to re-think an re-work that. But I spent a lot of time to investigate and came to the conclusion that it's not possible and I need to use Modal for the case where no resources can be deleted.


@CalamityC 
Apologies, the first one is missing `<Layout>`, that's my bad... like I said though not an issue



The second one is 100% fair enough. In general though if you've spent a long time investigating how to do something and come to a conclusion about why you have to do something in a way that isn't your first instinct, then I'd like to see a comment in the code. 


It means that the reviewer gets an insight into why the code might look different, and they can also verify that there's not some strange workaround somewhere, and then when the next developer comes to look at the code and has the same question they have something to start with as to why the code is written like that.





That said, this is now passing tests over the threshold, and I already approved the code itself so I shall merge :D

PR #1427, opened 2025-07-03
ERM-3751: JSONPath dependency

Jack-Golding:
Added JSONPath dependency to ui-agrements to assist with ERM-3697 work

ERM-3751

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
     "core-js": "^3.6.1",
     "eslint": "^7.32.0",
     "graphql": "^16.0.0",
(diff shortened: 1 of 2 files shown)


dcrossleyau:
This caused the Platform hourly builds to fail. See the Slack thread at "folio-hosted-reference-envs" channel. Pinned to the previous version via [platform-complete/pull/3429](https://github.com/folio-org/platform-complete/pull/3429) for snapshot branch.

Cc @Jack-Golding and @EthanFreestone

PR #1435, opened 2025-07-17
ERM-3753 Build filter component transformation

CalamityC:
* add filter component
* add utilities function
* add filterString manually to request URL
* add translations
* add GokbRoute test


src/components/utilities/getFilterConfig.js
```diff
@@ -0,0 +1,40 @@
+import { FormattedMessage } from 'react-intl';
+import gokbConfig from '../../../docs/gokb-search-v1';
+
+const getFilterConfig = () => {
```
Jack-Golding, line 4: This has been previously discussed but just documenting it here

Like with the work done within your results column config, we should have a recursive case switch statement to be used when handling type

src/routes/GokbRoute/GokbRoute.js
```diff
@@ -23,14 +29,23 @@ const GokbRoute = ({ location }) => {
...
-    return template({
+    const filterString = transformFilterString(query?.filters);
+
+    const baseQuery = template({
       input: query?.query || '',
       perPage: params?.perPage,
       offset,
     });
+
+    // Append `filterString` manually not possible with handlebar template
```
Jack-Golding, line 43: Happy with alot of this, already discussed moving the filter string handling into Monirehs query transform function when the search functionality is merged

Jack-Golding (changes requested):
We've already discussed the changes required for this namely moving the config handler to a recursive switch case statement similar to what has been done within the results columns work

Other than that, looking good

PR #1455, opened 2025-09-10
ERM-3782: Modify Actions menus in agreement lines

MonireRasouli:


EthanFreestone (changes requested):
Sorry, thisisn't what is being asked for.


This will disable the button if there are NO policies, not if there are restricting policies.

Please check the equivalent work in the agreements action menu to see how it decides whether or not to display an edit/delete button.


The fix to "canEditLoading -> canDeleteLoading" is correct though and can stay

PR #1459, opened 2025-09-16
Erm 3808

MonireRasouli:


MonireRasouli:
I wrapped the AgreementCreatRoute component in <CalloutContext.Provider> so that useContext(CalloutContext) has a valid sendCallout function during tests. Without it, the callouts wouldn’t trigger, and I couldn’t verify callout notification.

PR #1461, opened 2025-09-18
ERM-3773 Add display of agreements for resource

CalamityC:
* add config entry, registry function for agreementsList
* write wrapper component to work around endless loop
* write render component
* add translation
* add handler-stripes-registry


EthanFreestone:
This is the list of agreements for a GOKB Title right? I'd prefer not to create a new Registry resource for that, it can live under Gokb... maybe gokb.agreementsForResource


The other pattern I'd rather keep is that Registry components be set up in the registry directory. I understand that the remoteKB stuff is separated, but that's not importing the Reg component anyway, it's simply invoking the registry. I'd like to keep everything the registry is doing together (and add tests for the new component)

CalamityC:
> This is the list of agreements for a GOKB Title right? I'd prefer not to create a new Registry resource for that, it can live under Gokb... maybe gokb.agreementsForResource
> 
> The other pattern I'd rather keep is that Registry components be set up in the registry directory. I understand that the remoteKB stuff is separated, but that's not importing the Reg component anyway, it's simply invoking the registry. I'd like to keep everything the registry is doing together (and add tests for the new component)

First, there is no such gokb registered, it says explicitely gokbTIPPResource , which isn't in my work. Also, the component I wrote can be used for any remote kb, not only gokb, so I'd rather NOT name it gokb. And I do not understand what you mean with `gokb.agreementsForResource` - instead of `remoteKbAgreements`? But as I said: That component and therefore also that function can be used for any remote kb. Naming it after gokb seems wrong.

```
  const remoteKbAgreementsReg = registry.registerResource('remoteKbAgreements');
  remoteKbAgreementsReg.setRenderFunction('agreementsList', ({ remoteId, setBadgeCount }) => <RemoteResourceAgreementsList remoteId={remoteId} setBadgeCount={setBadgeCount} />);
```

And Second, sure I'll move the component to the registry directory

PR #1547, opened 2026-05-22
ERM-3992: On linking an eholding resource to an agreement line the link to view resource is incorrect

Jack-Golding:
Due to the EResourceLink component expecting an ERM shape, but has use within the agreement lines form where eholdings objects are used, ended up linking to an incorrect location


EthanFreestone (approved):
I think this has the potential to end up causing us issues, but I'll let it through for now.


So long as it doesn't end up directing our internal resources to a broken eholdings link I'm fine with this for now, and it's well commented

PR #1548, opened 2026-05-27
feat: Populate `resourceName` when adding an eHoldings resource to an agreement line via "eHoldings" -- ERM-4008

usavkov-epam:
## Purpose
https://folio-org.atlassian.net/browse/ERM-4007

Populate resourceName when linking an eHoldings resource to an agreement line so that it can be used for search, display

## Screen recordings

https://github.com/user-attachments/assets/e1c68937-5a7f-475a-a05d-d46de98c5c8c

https://github.com/user-attachments/assets/10c00dc4-a48e-4336-bd1e-e00894899b29

### DIFF
--- a/src/hooks/useExternalEntitlements/useExternalEntitlements.js
+++ b/src/hooks/useExternalEntitlements/useExternalEntitlements.js
@@ -1,0 +1,23 @@
+import { useQuery } from 'react-query';
+
+import { useOkapiKy } from '@folio/stripes/core';
+
+import { AGREEMENT_LINES_EXTERNAL_ENDPOINT } from '../../constants';
+
+export const useExternalEntitlements = (options = {}) => {
+  const {
+    enabled = true,
+    searchParams,
+    tenantId,
+  } = options;
+
+  const ky = useOkapiKy({ tenant: tenantId });
+
+  const result = useQuery({
+    queryKey: [AGREEMENT_LINES_EXTERNAL_ENDPOINT, searchParams, tenantId],
+    queryFn: () => ky.get(AGREEMENT_LINES_EXTERNAL_ENDPOINT, { searchParams }).json(),
+    enabled,
+  });
+
+  return result;
+};
--- a/src/hooks/useExternalEntitlements/useExternalEntitlements.js
+++ b/src/hooks/useExternalEntitlements/useExternalEntitlements.js
@@ -4,20 +4,31 @@ import { useOkapiKy } from '@folio/stripes/core';
 
 import { AGREEMENT_LINES_EXTERNAL_ENDPOINT } from '../../constants';
 
-export const useExternalEntitlements = (options = {}) => {
-  const {
-    enabled = true,
-    searchParams,
-    tenantId,
-  } = options;
+const defaultQueryNamespaceGenerator = (searchParams, tenantId) => [
+  'ERM',
+  'AgreementLine',
+  'external',
+  AGREEMENT_LINES_EXTERNAL_ENDPOINT,
+  searchParams,
+  tenantId,
+];
 
+export const useExternalEntitlements = ({
+  queryNamespaceGenerator = defaultQueryNamespaceGenerator,
+  queryOptions = {},
+  searchParams,
+  tenantId,
+} = {}) => {
   const ky = useOkapiKy({ tenant: tenantId });
 
-  const result = useQuery({
-    queryKey: [AGREEMENT_LINES_EXTERNAL_ENDPOINT, searchParams, tenantId],
-    queryFn: () => ky.get(AGREEMENT_LINES_EXTERNAL_ENDPOINT, { searchParams }).json(),
-    enabled,
+  const { data, isLoading } = useQuery({
+    queryKey: queryNamespaceGenerator(searchParams, tenantId),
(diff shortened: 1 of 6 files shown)


src/hooks/useExternalEntitlements/useExternalEntitlements.js
```diff
@@ -0,0 +1,23 @@
...
+  const {
+    enabled = true,
+    searchParams,
+    tenantId,
+  } = options;
+
+  const ky = useOkapiKy({ tenant: tenantId });
+
+  const result = useQuery({
+    queryKey: [AGREEMENT_LINES_EXTERNAL_ENDPOINT, searchParams, tenantId],
```
EthanFreestone: This is _not_ the pattern we use for queryKeys.

QueryKeys are hierarchical, so we namespace with `['ERM', 'AgreeementLine', 'external', ...`. This is so that we can refresh ALL agreement lines, or all external agreement lines, or everything in ERM, etc

src/hooks/useExternalEntitlements/useExternalEntitlements.js
```diff
@@ -0,0 +1,23 @@
+import { useQuery } from 'react-query';
+
+import { useOkapiKy } from '@folio/stripes/core';
+
+import { AGREEMENT_LINES_EXTERNAL_ENDPOINT } from '../../constants';
+
+export const useExternalEntitlements = (options = {}) => {
+  const {
```
EthanFreestone: Why isn't this directly destructured in the params?

src/hooks/useExternalEntitlements/useExternalEntitlements.test.js
```diff
@@ -0,0 +1,48 @@
...
+} from '@folio/jest-config-stripes/testing-library/react';
+import { useOkapiKy } from '@folio/stripes/core';
+
+import { externalEntitlements } from '../../../test/jest/entitlements';
+import { AGREEMENT_LINES_EXTERNAL_ENDPOINT } from '../../constants';
+import { useExternalEntitlements } from './useExternalEntitlements';
+
+jest.unmock('react-query');
+
+const queryClient = new QueryClient();
```
EthanFreestone: No. We already have a wrapper which provider query client for our tests, and patterns for making use of reactc-query in jest tests.

renderWithIntl will already provide a query client, and we have examples such as here:
https://github.com/folio-org/stripes-erm-components/blob/a2c4700f7f842d61e02832aee22ed50be908f3b5/lib/AccessControl/hooks/useCanAccess/useCanAccess.test.js#L1-L123


I don't mind exactly how useQuery is mocked per case, but we already handle all of this, I don't want individual tests to do special work

src/hooks/useExternalEntitlements/useExternalEntitlements.js
```diff
@@ -0,0 +1,23 @@
...
+
+  const ky = useOkapiKy({ tenant: tenantId });
+
+  const result = useQuery({
+    queryKey: [AGREEMENT_LINES_EXTERNAL_ENDPOINT, searchParams, tenantId],
+    queryFn: () => ky.get(AGREEMENT_LINES_EXTERNAL_ENDPOINT, { searchParams }).json(),
+    enabled,
+  });
+
+  return result;
```
EthanFreestone: We are in the middle of exploring a new pattern for individual fetch hooks.

I no longer want us to pin our hook API to react-query return API. If wee decide we need to change implementation, say to parallel batch fetch or to chunk the fetches, or later to replace react-query with another fetching library, even breaking changes in react-query itself, I don't want to have to go and change the implementation of everywhere making use of each hook.

To that end, we should explicitly surface all the properties we need.

I am _not_ against the raw return being provided for debugging andd/or edge cases, but it should be made clear in comments and/or README for the hook that use of the raw return is not guaranteed, and breaking changes can be made there, and that genuine use cases should surface properties they use explicitly

EthanFreestone (changes requested):
Some changes please. I'm fine with the approach at large, but I want to set good examples for this work, and in particular the test needs to follow our test structure and setup.

src/hooks/useExternalEntitlements/useExternalEntitlements.js
usavkov-epam: Thanks for the feedback, Ethan! That makes sense — I’ve updated the hook.

src/hooks/useExternalEntitlements/useExternalEntitlements.js
usavkov-epam: It’s mostly a matter of habit and readability 🙂 I’ve updated it to use inline destructuring without creating constants.

src/hooks/useExternalEntitlements/useExternalEntitlements.test.js
usavkov-epam: Updated

src/hooks/useExternalEntitlements/useExternalEntitlements.js
usavkov-epam: Updated

BogdanDenis (approved):

EthanFreestone (approved):
Thank you!

PR #1549, opened 2026-05-27
feat: Populate `resourceName` when adding an eHoldings resource to anagreement line via "Agreements" -- ERM-4008

usavkov-epam:
## Purpose

https://folio-org.atlassian.net/browse/ERM-4008

Populate resourceName when linking an eHoldings resource to an agreement line so that it can be used for search, display

## Screen recording

https://github.com/user-attachments/assets/f2693193-6952-4422-98e6-6dcc4705f4f8


src/routes/AgreementLineCreateRoute/AgreementLineCreateRoute.js
```diff
@@ -78,6 +78,7 @@ const AgreementLineCreateRoute = ({
         'type': 'external',
         'authority': resource?.type === 'packages' ? 'ekb-package' : 'ekb-title',
         'reference': resource.id,
+        resourceName: resource.name,
```
zburke, line 81: You don't need 'quotes' around the 'resourceName' key but maybe follow suit with all the other 'keys' that also don't need them? OCD like a boss, yo.

zburke (approved):

PR #1553, opened 2026-06-03
feat: ERM-4007 Add 'resourceName' to filters keys

usavkov-epam:
After #1549 #1548 are merged, we should add `resourceName` as a search key for the agreement lines.

https://folio-org.atlassian.net/browse/ERM-4007
https://folio-org.atlassian.net/browse/ERM-4008


SerhiiNosko (approved):

usavkov-epam:
Will be implemented in https://folio-org.atlassian.net/browse/ERM-4020

PR #1560, opened 2026-06-05
feat: Elegantly handle `404` from eHoldings to inform user of issue and display `resourceName` property for agreement lines -- ERM-4019

usavkov-epam:
## Purpose

https://folio-org.atlassian.net/browse/ERM-4019

For an EXTERNAL entitlement with a eHoldings reference, in there is an attempt to retrieve the resource data from the eHoldings API which fails (404) then the entitlement resourceName should be displayed (if populated) alongside information indicating there has been an issue retrieving the data from eHoldings (which could indicate some temporary issue, or could mean that the resource has been permanently removed from the underlying knowledgebase).

## Screen recording

https://github.com/user-attachments/assets/91f3aace-5f78-4400-bbcd-211bcaf96921


src/components/AgreementSections/LinesList/getAgreementLinesListFormatter.js
```diff
@@ -0,0 +1,132 @@
+import { FormattedMessage } from 'react-intl';
```
Jack-Golding: Whilst I understand why the decision was made to separate out the `agreementLinesFormatter` from the `LinesList` component, we usually reserve the centralisation of functions, components etc. for when they are used in more than one location within the codebase.

Additionally when this separation is also incorporating fixes for specific bugs, it results in the diffs being very difficult to parse and understand what changes were actually made.

For this change I would prefer if the formatter function (alongside associated `render*` functions) remained within the `LinesList` component and refactors with regards to file structure were left for potential future work

Jack-Golding (changes requested):
Other than the centralisation of the `getAgreementListFormatter` function mentioned in the comment, the PR looks like its on the right track. Updates to the tests, resources etc. are inline with patterns we have elsewhere

Once the requested change has been resolved I believe this PR should be good to go

src/components/AgreementSections/LinesList/getAgreementLinesListFormatter.js
usavkov-epam: Thank you for the feedback. I've updated the code based on your comment. Could you please take a look again?

src/components/AgreementSections/LinesList/LinesList.js
```diff
@@ -46,6 +46,20 @@ const columnMapping = {
   isCustomCoverage: ' ',
 };
 
+const hasReferenceError = (line) => !!line?.reference_object?.error;
+
+const wrapError = (renderFunc) => (agreementLine) => {
+  if (hasReferenceError(agreementLine)) return <></>;
```
EthanFreestone: Why can't this just be inline? It's a very simple boolean check, ~~and the fact that it's not set up as a callback means it's reinstantiated on render anyway.~~

(edit) Ah I see that this is set up outside of the component, that's not really necessary tbh, I'd prefer things like this to be set up as callbacks inside the component for consistency

EthanFreestone: I don't really like this pattern anyway, passing a function in as a function parameter is obfuscating what this code is really doing, which is effectively

line => !line?.reference_object?.error ? doTheThing : <></>


If you really really don't want to repeat this logic, I'd do it instead with a render component


```
const` LineErrorHandler = useCallback(({ children, line}) => {
  if (!!line?.reference_object?.error) {
    return; // I don't think you need need to return a node here tbh, but if you do then ovs <></> is ok
   }


   return children
}, [])
```



That way your pattern becomes 


```
line => (
  <LineErrorHandler line={line} >
     ......
  </LineErrorHandler>
)
```


Whilst this is _obviously_ more code, I'd argue that it's easier to read than passing a function as a parameter to another function all to return another render function.



Personally I'd just have the logic repeated, DRY is for not repeating big chunks of logic, not tiny boolean checks. At most maybe 


`const hasReferenceError = useCallback((line) => !!line?.reference_object?.error, [])`
and then

```
line => hasReferenceError(line) && doSomething
```

EthanFreestone (changes requested):
A couple of changes I'd suggest here, really just cleanup and making sure that the utility functions aren't completely obfuscating what's happening.

With such small changes as these DRY can often get in the way of actually readabl code. What's clear to you as the developer now might not be in 6, 8 months time, or 3 or 4 years.


It's often better to keep it as readable as possible, so balancing that with DRY

src/components/AgreementSections/LinesList/LinesList.js
usavkov-epam: Thanks for the suggestion @EthanFreestone, I see the intent behind moving this into a component-based guard.

I’m a bit hesitant about introducing a nested React component (especially with `useCallback`) in this case. From both a React and tooling perspective, this pattern is generally discouraged:

* SonarJS RSPEC-6478
* eslint-plugin-react `react/no-unstable-nested-components`
* eslint-react `no-nested-components`

In this case, the logic is purely a render guard (`if reference_object.error → skip render`) and doesn’t rely on any React-specific behavior (state, lifecycle, memoization, etc.). Because of that, I don’t really see a meaningful benefit in elevating it to a component abstraction compared to a simple function helper.

From my perspective, a function wrapper keeps the intent clearer and avoids introducing an additional React abstraction layer for something that is effectively a domain-level rendering rule.

Given that this condition is reused across multiple column renderers, I also think keeping it DRY here is justified — duplicating the same check in each renderer would reduce consistency and make future changes harder to maintain.

If we want to improve readability further, I’d suggest either:

* renaming the helper to something more expressive, e.g. `renderIfNoReferenceError`
* or adding a short inline comment describing the guard behavior

Example:

```js
// Do not render column if reference object is in error state
```

EthanFreestone: @usavkov-epam 

I understand, and I lost confidence in the nested component as I went alone, but I do think that my final suggestion balances the readability more nicely.

Function wraps are inherently more difficult to parse, and even if it's marginally the better call here it's a pattern we don't employ for that very reason, and so we start to drift code consistency


The last suggestion I made, namely

```
const hasReferenceError = useCallback((line) => !!line?.reference_object?.error, [])
 ...
line => hasReferenceError(line) && doSomething
```

perhaps is the nice intersection, keeping it DRY whilst being very readible and consistent with our patterns elsewhere. This very much is not an unstable component, and is no more nested than a function wrapper afaia


If there is a _major_ problem with this approach I haven't spotted then I can yield, but I just honestly don't think it's worth the loss of consistency and readibility.

EthanFreestone: In particular note how there are already functions being passed, some of which internally render further <IfPermission> components... 


My concern is mixing every approach together and a new junior developer coming in and having to parse through it all. Keeping it as a simple boolean check negates that and can way more easily be treated as a black box in an if/else.


I know it's a compromise, but the patterns when passing to MCLs could probably use a review in totality and that's definitely out of scope for this change

usavkov-epam: I've updated the implementation to use the simpler inline checks approach for consistency with the existing patterns in this area.

That said, I still believe the wrapper approach was a reasonable abstraction here, as it centralizes the repeated guard logic and keeps the individual formatters focused on rendering concerns. However, I agree that consistency within the current codebase is also an important consideration, so I've gone with the simpler version for this change.

@EthanFreestone @Jack-Golding, could you please take a look one more time?

EthanFreestone (approved):

PR #1561, opened 2026-06-08
feat: Support search by entitlement `resourceName` property in agreement lines search -- ERM-4020

usavkov-epam:
## Purpose
https://folio-org.atlassian.net/browse/ERM-4020

Use the entitlement `resourceName` for search and display when searching Agreements lines.


Jack-Golding (approved):
LGTM

PR #1562, opened 2026-06-24
ERM-4107, Added translation for Local KB resource name

MonireRasouli:


MonireRasouli:
added translation in the dashboard app instead of agreements

PR #1563, opened 2026-06-25
feat: Extend sort in agreement lines search and filter results to include by `resourceName` and `id` -- ERM-4103

usavkov-epam:
## Purpose
https://folio-org.atlassian.net/browse/ERM-4103

Currently when there are agreement lines results in the agreement line search and sort view, they are sorted by (in order):
- resource.name
- reference

This should be extended to include resourceName property as follows:
- resource.name
- resourceName
- reference
- id

This will ensure that lines sort in a consistent order no matter what properties are populated for the line


src/routes/AgreementLinesRoute/AgreementLinesRoute.js
```diff
@@ -80,7 +94,7 @@ const AgreementLinesRoute = ({
       page: currentPage,
       perPage: RESULT_COUNT_INCREMENT_MEDIUM,
       sortKeys: {
-        name: 'resource.name,reference',
+        name: SORT_KEYS.join(),
```
EthanFreestone: This is way overkill, we can just have the strings in the config here, it makes this significantly more readable.

I understand the desire to move to constants, but if we really wanted to do that I'd do it something like `AGREEMENTLINE.RESOURCE_DOT_NAME, AGREEMENTLINE.RESOURCE_NAME, AGREEMENTLINE.REFERENCE, AGREEMENTLINE.ID` , so at least they're still readable where they are actually configured.


However I'd suggest that that's a pattern we would want to either implement _throughout_ or nowhere (ie also using those for the keys in the MCL, or fetchers in the view pane, etc etc etc.


For consistency and ease of maintenance I suggest we simply leave these as strings as they are elsewhere in the codebase.


I think this may be worth revisiting in the event that we move the code over to Typescript, to ensure we're hardening against properties that exist on given domain classes

usavkov-epam: I see your point. I've updated that place in the code.

